PR #12, opened 2015-01-16
Test coverage and coveralls report.

jimhester:
Adds an automatic test coverage report using coveralls.

You will need to enable coveralls for this repository before merging this pull for it to work the first time.

https://coveralls.io/repos/new

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,10 @@ before_install:
   - chmod 755 ./travis-tool.sh
   - ./travis-tool.sh bootstrap
 
-#install:
+after_success:
+  - Rscript -e 'library(covr);coveralls()'
+install:
+  - ./travis-tool.sh github_package jimhester/covr
 #  - ./travis-tool.sh install_r rbenchmark microbenchmark
 
 script: ./travis-tool.sh run_tests
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 ## digest [![Build Status](https://travis-ci.org/eddelbuettel/digest.png)](https://travis-ci.org/eddelbuettel/digest)
-[![Coverage Status](https://img.shields.io/coveralls/jimhester/digest.svg)](https://coveralls.io/r/jimhester/digest?branch=master)
+[![Coverage Status](https://img.shields.io/coveralls/eddelbuettel/digest.svg)](https://coveralls.io/r/eddelbuettel/digest?branch=master)
 
 ## What
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 ## digest [![Build Status](https://travis-ci.org/eddelbuettel/digest.png)](https://travis-ci.org/eddelbuettel/digest)
+[![Coverage Status](https://img.shields.io/coveralls/jimhester/digest.svg)](https://coveralls.io/r/jimhester/digest?branch=master)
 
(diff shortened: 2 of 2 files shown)


eddelbuettel:
Thanks for the PR. I watched your work with interest after you pinged me in that thread.

I have not yet had time to research this, and have not formed an opinion as to whether I want to enable this in my packages or not.  So for now I'll let this (unsolicited) pull request rest.

jimhester:
Sounds good to me,  https://coveralls.io/jobs/4051949 has an example report for digest if you wanted to see what the report would look.  (I was testing compiled coverage with it).

eddelbuettel:
Yes, thanks, I saw the report. It is nicely detailed and color-coded.  What is the added run-time for a small package like digest?

jimhester:
Looks like 8 seconds total for digest
https://travis-ci.org/jimhester/digest/builds/47294358#L735

It has to recompile the code with the tracing instrumentation turned on and
with optimizations off, then run all the tests.   I am not sure of the
breakdown between compile time and run time in the above 8 seconds.  I
would expect most of it was compile time.

On Fri, Jan 16, 2015 at 4:27 PM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
That sounds a little low.  A full R CMD ... cycle takes longer.  Is it just build and test?

jimhester:
It uses `devtools::load_all(recompile = TRUE)` to compile the code, so it doesn't have to do a full `R CMD build`, which is why it is faster.

Also most of that 8 seconds is likely from the R interpreter starting and transferring the json file to coveralls.io.

On my machine from an already running R
```r
system.time(package_coverage('digest'))
#  user  system elapsed
#  1.327   0.250   1.626

# Just running the tests takes basically no time however
system.time(testthat::source_dir('digest/tests/'))
#    user  system elapsed
#  0.028   0.003   0.033
```

eddelbuettel:
Let me rephrase then:

What is time difference in run-time ("all-in") for a Travis build with this change, versus no change.  

This would have to account for installing devtools as its dependencies too as my packages work without devtools (but maybe r-travis changed again and forces it in now, which is possible).

jimhester:
from https://travis-ci.org/eddelbuettel/digest/builds/47294622#L406 it looks like installing `covr` and dependencies took an additional 27 seconds, although that seems to depend greatly on what machine is running the build, https://travis-ci.org/jimhester/digest/builds/47294207#L399 only took 19 seconds and https://travis-ci.org/jimhester/digest/builds/46648160#L399 only took 15 for instance.

It looks to me like it adds about 30 seconds to the build time overall.

eddelbuettel:
That's not so bad. I often spend more time re-installing other build dependencies.

Would be nice to _finally_ be able to supply custom Docker containers with covr in them...

eddelbuettel:
I (finally !!) merged this, but I had to do it by hand, which sadly does not show you as author.  Sorry about that.  Thanks for your patience with this.

See https://travis-ci.org/eddelbuettel/digest/builds/179057656 -- it barks a little at the about loading digest when digest is already loaded.  I have not looked closely but I suspect it goes back to covr loading digest?

eddelbuettel:
Closing as just discussed.  Thanks for the PR.  It is (at long last) in.

PR #17, opened 2015-10-11
fix #16

### LINKED ISSUE #16: Long vector support relied on, but not available
Processing a large R object (that otherwise behaves well with most manipulations in R) results in an error with `digest`.  Specifically:
```
Error in digest(user_action, algo = "md5") : 
  long vectors not supported yet: memory.c:3361
```
Nearly identical errors (just the call is different) occur for other algorithms.  Using debug I can see that the error is triggered at `val <- .Call(digest_impl, object, as.integer(algoint), as.integer(length), as.integer(skip), as.integer(raw), as.integer(seed))`; which I believe either places the error in `.Call` or in `digest`.  Memory.c appears to be outside of the scope of `digest` but belongs to R instead.  Nevertheless, because I'm able to work with and manipulate this object in every other context it makes me think that some step along the way in `digest` is the root of the problem.

Example data:
```
eg <- data.frame(V1=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000),V2=runif(100000000),V3=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000),V4=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000),V5=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000))
digest(eg)
```

Session Info:
```
R version 3.2.2 (2015-08-14)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 14.04.3 LTS

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] digest_0.6.8

loaded via a namespace (and not attached):
[1] tools_3.2.2
```

Perhaps something else could be done in the digest internals to make it not rely on memory.c or an overly large object could be detected, truncated, and a hash returned with a warning?

thirdwing:
To fix #16 . Just replace `LENGTH`with `XLENGTH`.

```r
> library(digest)
> eg <- data.frame(V1=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000),V2=runif(100000000),V3=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000),V4=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000),V5=rep("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",100000000))
> digest(eg)
[1] "d41d8cd98f00b204e9800998ecf8427e"
```

### DIFF
--- a/src/digest.c
+++ b/src/digest.c
@@ -58,7 +58,7 @@ SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Se
     int output_length = -1;
     if (IS_RAW(Txt)) { /* Txt is either RAW */
         txt = (char*) RAW(Txt);
-        nChar = LENGTH(Txt);
+        nChar = XLENGTH(Txt);
     } else { /* or a string */
         txt = (char*) STRING_VALUE(Txt);
         nChar = strlen(txt);


eddelbuettel:
Seriously?  One char?  `LENGTH` to `XLENGTH`?  Nice :)

thirdwing:
Theoretically, there are many places to change, like https://github.com/eddelbuettel/digest/blob/master/src/raes.c#L30

But I don't think anyone will try to use that long key.

BTW, MetroHash looks very interesting.

thirdwing:
I am so sorry...

I need to be more careful. [This line](https://github.com/eddelbuettel/digest/blob/master/src/digest.c#L57) 

```
int nchar
```

should be changed into 

```
uint32 nchar
```

eddelbuettel:
Will do. We probably also need a reference case or two to test.

PR #20, opened 2015-12-27
calculate stable SHA1

ThierryO:
`get_sha1()` calculates stable SHA1 hashes: the hash on 32-bit and 64-bit systems is identical. This is ensured by applying `num_32_64()` on numerics. `sha1_digits()` defines the maximum number of digits to take into account when calculating the hash. If the number of digits is too high then the hash on 32-bit will differ from that on 64-bit systems. This is due to the difference in floating point precision.


eddelbuettel:
Hi Thierry,

Thanks for this.  It looks like a nice piece of work. In general, it may be preferable to first open issue tickets to discuss matters of style etc.  I see two departures here

i)  camelCase preferred over snake_case -- could I twist your arm to rename to `getSha1()` or even, sticking with the rest of the short names here, just `sha1()` ?
ii) I am not a huge fan of S4 and as this is pretty widely-used package I would prefer to keep the loadtime minimal.  Can you see a way to do what you did in S3 only?  Also, do the two helper functions need to be exported?
iii) More trivial:  four space indentation preferred as in the other files.

Thanks,  Dirk

ThierryO:
Hi Dirk,

Sorry for not opening an issue first. I've written the code for one of our projects and it took quite some work to get the 32 vs 64 bit thing working. I guess that other people can use it as well. I took the code from our project with minimal changes (removing Roxygen2 and converting testthat unit tests). Hence the differences in coding style.

1. I'd prefer short names likes `sha1()`, `sha1digits()`, `num3264()` over camelCase. CamelCase won't fit with our coding style ;-)
1. I'm more familiar with S4 than S3. However I will attempt to get it working with S3. 
1. I would export the helper functions. Because it's easier when defining methods for new classes. If `num3264()` isn't exported, you'll need to call `sha1()` recursively.
1. `sha1digits()` just sets some sensible defaults. Maybe it's better to put them as different objects in `sysdata.rda`.
1. Updating to four space indentation is not problem.

Thanks for considering the pull request.

Best regards,

Thierry

eddelbuettel:
Hi Thierry,

You rock.  Thanks for the quick reply.  Regarding the points in order:
1)  Nice.  I like all those better.  Maybe `num32to64` to split the digits apart?
2)  I'd appreciate it.  If not, well, it is a small addition -- and I realized that just adding the `methods` package is not that big a deal.  I'd take it as is; if you can make work in S3 I'd be even happier.
3)  Ok.  And you already wrote the help texts :)
4)  Could do. Or even in `options()` ?
5)  ESS is really good at that too :)

Looking forward to an update!

Dirk

PS One minor niggle also is that `DESCRIPTION::Version` should probably just be 0.6.8.4 so that we room for an iteration or two before we call it 0.6.9 (or even 0.7.0).

ThierryO:
Hi Dirk,

I think I've resolved your requests. The changes are currently in the [S3 branch of my fork ](https://github.com/ThierryO/digest/tree/S3). Comments on the changes are welcome.

Before merging S3 into master, I'll test this in our private package on two points: a) does `sha1()` work when imported in another package. b) can one add additional S3 methods for `sha1()` in another package.

Thierry

eddelbuettel:
Thierry,

Outstanding. Just glanced at your commit log titles.  Nice work.  

I guess you have nice examples too for 'digest'-ing numerical etc input this way?

Dirk

eddelbuettel:
Salut --  And Happy New Year!   Any news on your tests?  Is this good to go now?

ThierryO:
A Happy New Year!. Yes the tests worked out fine. I've merged the S3 branch into the pull request.

There is one big change: sha1.default() returns an error instead of using digest(). This ensures that sha1() is only used when a method is available. Otherwise the user thinks that he gets a stable hash when it's not.

Due to this change I might need to add some more methods. Maybe one for every class in base R? Shall I do that first or do it in a seperate PR so that you can merge this one?

R/sha1.R
```diff
@@ -0,0 +1,177 @@
+# functions written by Thierry Onkelinx
+sha1 <- function(x, digits = 14, zapsmall = 7){
+    UseMethod("sha1")
+}
+
+sha1.default <- function(x, digits = 14, zapsmall = 7) {
+    stop("sha1() has not method for the '", class(x), "' class")
```
eddelbuettel: Minor rewording:  _has no method_.  But we get that next time.

Also, I may consider adding `call.=FALSE` -- but I don't do this yet in other places in digest so no rush.

eddelbuettel:
I'd say new PR.

PR #24, opened 2016-01-05
bugfix: Trailing zeros in num2hex()

ThierryO:
`num2hex()` removed only the last trailing zero instead of all trailing zeros. `digest:::num2hex(sqrt(2) ^ 2)` returned `"00000000000 1"` instead of `" 1"` which is the output of `digest:::num2hex(2)` As a result `sha1(2) != sha1(sqrt(2) ^ 2)`. This is now fixed.

### DIFF
--- a/R/sha1.R
+++ b/R/sha1.R
@@ -170,7 +170,7 @@ num2hex <- function(x, digits = 14, zapsmall = 7){
   mantissa <- gsub(mantissa, pattern = "p.*$", replacement = "")
   mantissa <- substring(mantissa, 1, digits.hex) # select required precision
   # remove potential trailing zero's
-  mantissa <- gsub(mantissa, pattern = "0$", replacement = "")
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Nice catch.

BTW how do you feel about calling a release before we have a vignette?  Yes/No/Maybe/Don'tCare?

eddelbuettel:
Also, any chance I could get to fill in the [ChangeLog](https://github.com/eddelbuettel/digest/blob/master/ChangeLog) ?  Standard GNU format and autoformatting in Emacs ...

ThierryO:
I'd release it as is. And release it again as soon as the vignette is ready.

I'll have a go at the ChangeLog. I hope to get the formatting right as I don't use Emacs.

eddelbuettel:
100% agreed to release. Will get to that this eve,

Never to late to learn Emacs :)

PR #25, opened 2016-01-06
update ChangeLog with the additions for sha1()

ThierryO:


eddelbuettel:
Looks good.  

Wasn't sure if you wanted it spread over several dates.  It really is just one logica block so ...

Issue #27, opened 2016-01-10
sha1() fails if x contains an empty list


mllg:
Small example:
```r
sha1(list())
sha1(list(a = 1, b = list()))
```

eddelbuettel:
Ooops.  Has an element of pilot error but maybe we should test more.

@ThierryO do you want to take a stab at this?

ThierryO:
I'll have a look at it

ThierryO:
PR #28 should solve the problem.

eddelbuettel:
Closing as #28 has been merged.

ThierryO:
This issue is still open ;-)

eddelbuettel:
PRs welcome.

vsimko:
```r
sha1(list())
sha1(list(a = 1, b = list()))
```
... both do work, should this issue be close already ?

eddelbuettel:
Maybe. Any consensus view?  Any view by @mllg?

ThierryO:
IMHO this is resolved. You can close the issue.

eddelbuettel:
Will do.

PR #28, opened 2016-01-10
use vapply instead of the error-prone sapply

mllg:
Fix for #27.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2016-01-11 Thierry Onkelinx  <thierry.onkelinx@inbo.be>
+
+	* R/sha1.R: handle empty list and empty dataframe (#issue 27)
+	* R/sha1.R: take the object class, digits and zapsmall into account (#PR 28)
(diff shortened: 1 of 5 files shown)


mllg:
On a second thought, my commit is not very clever...
```r
all.equal(sha1(list()), sha1(character(0)))
```

mllg:
The previous issue is fixed. However, I'm curious if this is supposed to happen:
```r
# TRUE on CRAN and patched
identical(sha1(list(a = 1)), sha1(data.frame(a = 1)))
```
Same issue for other R objects with attributes.

ThierryO:
Thanks for the PR. I'll add some more stuff.

ThierryO:
I've added my changes into a [PR at mllg:master](https://github.com/mllg/digest/pull/1).

Should `sha1()` take attributes of the original object into account? Now they are ignored except is the simple cases where `sha1()` is just a wrapper for `digest()`.

eddelbuettel:
Thank you both!   I'll also add Michel to the Authors list.

(And I now see why people like 'squashed commits' on PR but I don't mind explicit ones.)

Issue #23, opened 2016-01-04
Classes that sha1() should have methods for in digest


ThierryO:
- Date
- complex
- array

vsimko:
already implement
- `complex` converted to a 2-column matrix (Re, Im)
- `Date` converted to an integer (positive/negative) representing number of seconds since 1970

any ideas for the `array` type ?

eddelbuettel:
Same as for the other two?  There is `as.array()` though I am not sure what one usually converts from...

vsimko:
`array` is a multi-dimensional matrix. One of the problems is to produce different hashes for differently shaped arrays that contain the same data:
```r
# the same content for differenly shaped arrays
content <- 1:8

# creates a cube 2x2x2
a <- array(content, dim = c(2,2,2))
b <- array(content, dim = c(2,4,1))

# structures to be hashed
a_tohash <- list(
    dim = dim(a),
    data = as.numeric(a)
)

b_tohash <- list(
    dim = dim(b),
    data = as.numeric(b)
)

# should produce different hashes
sha1(a_tohash)
sha1(b_tohash)
```

vsimko:
any other problems that need to be addressed before I create a pull request ?

eddelbuettel:
Ah, yes, of course.   

Well a 1:8 sequence with dim(2,2,2) is different from dim(2,4,1) so I think this is good enough.

eddelbuettel:
Unit tests maybe (in the super-simple format we use here) ?

vsimko:
**Note:** the current implementation assumes that: `sha1(2+0i) != sha1(2)` (conservative approach, but maybe mathematically incorrect)

vsimko:
I've  just found a related section in the vignette titled **"Creating a sha1 method for other classes"**: https://github.com/eddelbuettel/digest/blob/master/vignettes/sha1.Rmd#creating-a-sha1-method-for-other-classes

Useful for new contributors.

vsimko:
Any types that should also be implemented ?

ThierryO:
I've started this issue so everyone could add other classes to the wishlist. Keep in mind that @eddelbuettel wants to keep the list of dependencies as minimal as possible. That limits the classes to those available in base R.

eddelbuettel:
Yes, I think anything besides base R is likely to be out of scope. Unless we all agree that it is worth it.

But minimal dependencies are a Good Thing (TM) for a widely-used utility package such as this.

[...]

PR #38, opened 2016-05-01
support for array + unit tests

vsimko:
- array encoded as a list(dim, data)
- added some unit tests
- This pull request fixes the whole issue #23


eddelbuettel:
Super -- home now after some traveling.

PR #41, opened 2016-05-25
Add sha1 method for pairlist

ThierryO:


eddelbuettel:
Looks good, will merge in due course.

PR #54, opened 2017-01-11
add sha1 method for the class "function"

ThierryO:


ThierryO:
That was lightning fast :-)

eddelbuettel:
Aligned with the beginning of the workday here :)

Issue #70, opened 2018-01-12
sha1() fails on empty matrices


ThierryO:
```
> sha1(matrix(integer(0)))
Error in x[1, 1] : subscript out of bounds
```
I'll look into this.

ThierryO:
Solved by #71

PR #71, opened 2018-01-12
Empty matrix

ThierryO:
Solves #70 

I took the liberty of moving the license file to a location where GitHub can find it. I'll revert that if you don't like it.

Updating the version number and ChangeLog is still to do.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2018-01-12  Thierry Onkelinx <thierry.onkelinx@inbo.be>
+
+    * sha1() handles empty matrices
+
 2017-12-20  Dirk Eddelbuettel  <edd@debian.org>
 
 	* src/raes.c (AESencryptECB,AESdecryptECD): Replace two uses of NAMED
--- a/R/sha1.R
+++ b/R/sha1.R
@@ -5,7 +5,7 @@ sha1 <- function(x, digits = 14L, zapsmall = 7L, ...){
 
 sha1.default <- function(x, digits = 14L, zapsmall = 7L, ...) {
     stop(
-        "sha1() has not method for the '",
+        "sha1() has no method for the '",
         paste(class(x), collapse = "', '"),
         "' class",
         call. = FALSE
--- a/R/sha1.R
+++ b/R/sha1.R
@@ -74,7 +74,7 @@ sha1.numeric <- function(x, digits = 14L, zapsmall = 7L, ...){
 
 sha1.matrix <- function(x, digits = 14L, zapsmall = 7L, ...){
     # needed to make results comparable between 32-bit and 64-bit
-    if (class(x[1, 1]) == "numeric") {
+    if (storage.mode(x) == "double") {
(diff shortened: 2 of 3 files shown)


DESCRIPTION
```diff
@@ -18,6 +18,6 @@ Description: Implementation of a function 'digest()' for the creation
 URL: http://dirk.eddelbuettel.com/code/digest.html
 BugReports: https://github.com/eddelbuettel/digest/issues
 Depends: R (>= 2.4.1)
-License: GPL (>= 2)
+License: GPL (>= 2) | file LICENSE
```
eddelbuettel: I don't think CRAN likes it that way.  I would revert that.

tests/digestTest.R
```diff
@@ -165,7 +165,7 @@ for (i in seq(along.with=murmur32Input)) {
 }
 ## test 'length' parameter and file input
 ##fname <- file.path(R.home(),"COPYING")  ## not invariant across OSs
-fname <- system.file("GPL-2", package="digest")
+fname <- system.file("LICENSE", package="digest")
```
eddelbuettel: Same here.

tests/digestTest.R
```diff
@@ -190,7 +190,7 @@ for (alg in c("sha1", "md5", "crc32", "sha256", "sha512",
 ## compare md5 algorithm to other tools
 library(tools)
 ##fname <- file.path(R.home(),"COPYING")  ## not invariant across OSs
-fname <- system.file("GPL-2", package="digest")
+fname <- system.file("LICENSE", package="digest")
```
eddelbuettel: And here.

eddelbuettel:
So if it is to have the _repo_ have a license to display, I would simply make a copy, or a link.  GitHub's file system does not care as it has clever de-duplication.  We then add the new location to `.Rbuildignore` and all is well.

What I am a little more concerned about it the large set of lines in the output that changes.  What caused that?  This seems a little disruptive.

And over all this PR is not clear: what were you trying to do in the first place?  I feel the GPL-2 file move/renamer obfuscated something.  What issues with empty matrices needed fixing / were fixed?   In sum, this may be a good example of why starting with an issue ticket is better....

eddelbuettel:
I see now re #70.  Absoliutely worth fixing, and the `storage.mode` change is good, as were the typos.

DESCRIPTION
ThierryO: It passes R CMD Check --as-cran. But I'll revert it.

ThierryO:
Shall I bump the package version to 0.6.13.2?

eddelbuettel:
Could do but don't have to.

The "why are there dozens of lines of digest output different" question is more important.  What drove that?

ThierryO:
I think this is ready to merge now

eddelbuettel:
(One totally minor thing I'll just fix on my end:  by some historic convention ChangeLog separates user name and email by two spaces. No need to update now.  I'll fix over morning coffee tomorrow.)

PR #90, opened 2018-08-31
This patch changes the SHA512 code to avoid faults from mis-aligned accesses

radfordneal:
In particular, such faults arise with R-3.5.1 and digest-0.6.16 on a
SPARC T2 system running Solaris 11 when compiled with gcc 4.9.2 in
32-bit mode, as illustrated below:

------------------------------------------------------------------------

R version 3.5.1 (2018-07-02) -- "Feather Spray"
Copyright (C) 2018 The R Foundation for Statistical Computing
Platform: sparc-sun-solaris2.11 (32-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> library(digest)
> txt <- ""
> for (i in 1:10) txt <- paste0(txt,"abcdefghijklmnopqrstuvwxyz")
> r <- digest(txt,alg="sha512",serialize=FALSE,skip=0)
> r <- digest(txt,alg="sha512",serialize=FALSE,skip=8)
> r <- digest(txt,alg="sha512",serialize=FALSE,skip=4)

 *** caught bus error ***
address 24cb07c, cause 'invalid alignment'

Traceback:
 1: digest(txt, alg = "sha512", serialize = FALSE, skip = 4)
An irrecoverable exception occurred. R is aborting now ...
Bus Error (core dumped)

------------------------------------------------------------------------

Whether the problem occurs varies with the compiler used, since some
SPARC instructions require alignment and others do not (so depends on
which instructions the compiler chooses).  The problem presumably also
occurs on other platforms with alignment requirements.  (Perhaps some
ARM systems?)

The fix is made to sha2.[ch], which is used for SHA512. There is an
analogous problem with the SHA256 code in those files, which is also
fixed (untested) in this patch, but it is in any case not enabled in
the 'digest' package (which uses other code for SHA256).

The issue is in the SHA512_Transform function, which is used locally
in sha2.c.  It is declared as

   void SHA512_Transform(SHA512_CTX* context, const sha2_word64* data)

Alignment faults may arise when it is called as

   SHA512_Transform(context, (sha2_word64*)data);

since 'data' may not be 64-bit aligned.  It is also called as

   SHA512_Transform(context, (sha2_word64*)context->buffer);

which does not produce a fault because context-buffer happens to be
64-bit aligned.  This call does, however, violate the 'const'
attribute in the prototype, since context->buffer is written to inside
SHA512_Transform.  (This aliasing violates the C standard, but is
unlikely to actually cause a bug.)

This patch modifies SHA512_Transform so it is declared as

   static void SHA512_Transform(SHA512_CTX* context)

with data always coming from context->buffer.  The previous call with
'data' as the second argument is replaced by

   MEMCPY_BCOPY(context->buffer, data, SHA256_BLOCK_LENGTH);
   SHA256_Transform(context);

The copy should handled any alignment for 'data'.

There is no measurable performance impact.

### DIFF
--- a/src/sha2.c
+++ b/src/sha2.c
@@ -29,6 +29,8 @@
  * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  *
+ * Modified by R. Neal, 2018, to avoid mis-alignment faults.
+ *
  */
 
 #include <string.h>	/* memcpy()/memset() or bcopy()/bzero() */
@@ -218,9 +220,9 @@ typedef u_int64_t sha2_word64;	/* Exactly 8 bytes */
  * library -- they are intended for private internal visibility/use
  * only.
  */
-void SHA512_Last(SHA512_CTX*);
-void SHA256_Transform(SHA256_CTX*, const sha2_word32*);
-void SHA512_Transform(SHA512_CTX*, const sha2_word64*);
+static void SHA512_Last(SHA512_CTX*);
+static void SHA256_Transform(SHA256_CTX*);
+static void SHA512_Transform(SHA512_CTX*);
 
 
 /* /\*** SHA-XYZ INITIAL HASH VALUES AND CONSTANTS ************************\/ */
@@ -350,7 +352,7 @@ void SHA256_Init(SHA256_CTX* context) { /* #nocov start */
 #if BYTE_ORDER == LITTLE_ENDIAN
 
 #define ROUND256_0_TO_15(a,b,c,d,e,f,g,h)	\
-	REVERSE32(*data++, W256[j]); \
+	REVERSE32(W256[j], W256[j]); \
 	T1 = (h) + Sigma1_256(e) + Ch((e), (f), (g)) + \
              K256[j] + W256[j]; \
 	(d) += T1; \
@@ -362,7 +364,7 @@ void SHA256_Init(SHA256_CTX* context) { /* #nocov start */
 
 #define ROUND256_0_TO_15(a,b,c,d,e,f,g,h)	\
 	T1 = (h) + Sigma1_256(e) + Ch((e), (f), (g)) + \
-	     K256[j] + (W256[j] = *data++); \
+	     K256[j] + W256[j]; \
 	(d) += T1; \
 	(h) = T1 + Sigma0_256(a) + Maj((a), (b), (c)); \
 	j++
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thanks a lot, @radfordneal, and apologies that this fell to the side for a few days.  

@hannesmuehleisen You were the one who added the `sha*` summaries, would you be able to take a look at this?  I am mildly concerned that some of the details are altered, but then again this is of course offset by the same that Radford is a very known and trusted source as well as the fact that the existing tests all pass (no proof, of course, but a feel-good factor).

Otherwise, would either one of you know of another implementation that passes muster on all little-vs-big endian, 32-vs-64 bit OS, ... cases?  I presume that that Sparc system Radford uses here is a little unusual as we have not heard from anyone else yet?

eddelbuettel:
No word from @hannesmuehleisen , sadly, but I think this should be fine.  

I'll try remember to release this to CRAN once CRAN reopens in a few days.

hannesmuehleisen:
Sorry only seeing this now. This is a great improvement, thanks, @radfordneal. As far as I'm concerned, as long as the tests and checks still pass, I'm happy.

eddelbuettel:
This is now on CRAN.  As a side effect, it also took out an UBSAN report against the file.

I still have a few in `xxhash.c` (see the reports at https://cloud.r-project.org/web/checks/check_results_digest.html) all of which point at 4 (or 8) byte alignment.  

If either of you, or @jimhester who contributed `xxhash.c` support, have spare cycles for this I would truly appreciate help with this.

eddelbuettel:
I took another look and I may have to tell CRAN that two remaining [UBSAN issues](https://cloud.r-project.org/web/checks/check_results_digest.html) (now updated after our newest version got onto CRAN with the changes by @radfordneal):
- all `#ifdef` for an alignment choice in `xxhash.c` lines 227 and 240
- in `pmurhash.c` line 276 is it hidden away in an upstream macro that has not changed

are possibly false positives.   Any comments?

radfordneal:
Regarding the note on line 276 of pmurhash.c, the problem is that the
macro used has the line

    c = c>>8 | *ptr++<<24;

where c is a uint32_t and ptr is a pointer to uint8_t.  Since *ptr++
will be of uint8_t type, it will
be subject to "integer promotion", in which the C99 standard says "If
an int can represent all
values of the original type, the value is converted to an int".  Once
converted to int (which
is 32 bits in size, and signed), a left shift of the value 254 results
in an overflow, which is
undefined behaviour.

I think that changing it to

   c = c>>8 | (uint32_t)*ptr++<<24;

would fix the problem.  In theory, *ptr++ will be converted first to
an int, and then to uint32_t,
before being shifted.  All these operations are guaranteed to have
defined behaviour.

It's unlikely that the problem would actually cause a bug, though it's
hard to be sure if the
compiler is overly-enthusiastic about assuming that undefined
behaviour can't occur.

     Radford

eddelbuettel:
From a first casual check -- of changing the one line after merging the @jimhester pr #91 -- it looks good!

Thanks again for the quick heads-up.

PR #91, opened 2018-09-13
Update to use xxHash v0.6.5

jimhester:
This should fix the alignment errors detected by UBSAN

I tested this in a rocker/r-devel-ubsan-clang docker container, and it passed cleanly, so I believe it should fix the issue on CRANs setup as well.

The diff is a bit bigger than just the xxhash files because the API changed a bit since it was first added.

cc https://github.com/eddelbuettel/digest/pull/90#issuecomment-421076421

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2018-09-02  Jim Hester  <james.f.hester@gmail.com>
+
+	* src/xxhash.c: Updated to use xxHash v0.6.5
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Nice one!  I had planned to look for an updated version but had not gotten to it.  So thanks for this.

eddelbuettel:
Leaves `pmurhash.{c,h}` which is also you and me ...

PR #93, opened 2018-10-29
add digest2int, fixes #82

### LINKED ISSUE #82: string -> int hashing
Hi here. I'm wondering whether it make sense to add simple hash function to hash arbitrary string into 32 bit integer. It seems it is quite common case and everyone reinvents the wheel (I'm opening this issue because I found myself reimplementing this 3rd time for different packages). 

One solution is to use `digest` with `Rmpfr::mpfr`. But it is another pkg and system dependency (`libmpfr`). 

`digest` itself is widely available and would be nice to have simpler built-in solution like [One-at-a-Time](https://en.wikipedia.org/wiki/Jenkins_hash_function#one_at_a_time) or something from this [article](http://eternallyconfuzzled.com/tuts/algorithms/jsw_tut_hashing.aspx). 

I found `One-at-a-Time` works nice - low collision and super fast. I can send PR.

dselivanov:


DESCRIPTION
```diff
@@ -4,8 +4,8 @@ Author: Dirk Eddelbuettel <edd@debian.org> with contributions
  Mario Frasca, Bryan Lewis, Murray Stokely, Hannes Muehleisen,
  Duncan Murdoch, Jim Hester, Wush Wu, Qiang Kou, Thierry Onkelinx, 
  Michel Lang, Viliam Simko, Kurt Hornik and Radford Neal.
-Version: 0.6.18
```
eddelbuettel: Dev versions get a minor first. Please make it 0.6.18.1

dselivanov: I will fix. Anything else apart from this?

eddelbuettel: No, it looks good, and I like it.

I am wondering if we should apply one lesson in 10+ years here and make it vectorised in and out, ie accept a vector of strings and return a vector of ints?  Thoughts?

dselivanov:
But it is already vectorized! :-) (loop at C level)

пн, 29 окт. 2018 г., 19:56 Dirk Eddelbuettel <notifications@github.com>:

eddelbuettel:
> But it is already vectorized! :-) (loop at C level)

Silly me. I was in a rush and looking for the Rcpp vector but it is of course a no-depends package :)

dselivanov:
@eddelbuettel if you want to make a new release could you please wait a little bit? I found that results from initial Rcpp implementation (from the #82 ticket) and merged C implementation are different for certain cases. I want to investigate why it that.

eddelbuettel:
Thanks for the heads-up. Will hold off then.

dselivanov:
The difference is due to these lines:
```c++
#ifndef uint32_t
#define uint32_t unsigned long int
#endif
```
On 64 bit architectures it is 64 bit unsigned integer... I'm noob in such things and always assumed `uint32_t` is available everywhere. Also I'm not sure why the macro triggered. If I comment out it, the file will still compile and `uint32_t` from `<stdint.h>` is available...

Any thoughts?

eddelbuettel:
Whoa.  That line looks like a blooper.

I am with you -- "these days" we have `(u)int(32|64)_t`.  In the bad old days people did hacks like the `#define` there ... which we could teach about Windoze but the better approach would be consistent use of `(u)int32_t`.   Can we do that?

(And that reminds me that I have another PR in another repo with something like that....)

dselivanov:
I will remove define and send PR

чт, 8 нояб. 2018 г., 20:07 Dirk Eddelbuettel notifications@github.com:

Issue #97, opened 2019-04-20
digest can't detect small change to large dataset


kendonB:
I came across this issue in real life using the workflow package [drake](https://ropensci.github.io/drake/) which calls `digest` via the storage package [storr](http://richfitz.github.io/storr/). I had made a moderate change to one column of a 35 column and 29906368 row dataset (5GB in memory) and drake did not detect the change.

```r
# make a big dataset
big_iris = do.call(dplyr::bind_rows, rep(list(iris), 800000))

# make a minor change
big_iris2 = big_iris
big_iris2$Species[500000]
#> [1] setosa
#> Levels: setosa versicolor virginica
big_iris2$Species[500001]
#> [1] versicolor
#> Levels: setosa versicolor virginica
tmp = big_iris2$Species[500000]
big_iris2$Species[500000] = big_iris2$Species[500001]
big_iris2$Species[500001] = tmp
identical(big_iris, big_iris2)
#> [1] FALSE

# digest thinks they're the same
digest::digest(big_iris2) == digest::digest(big_iris)
#> [1] TRUE
```

<sup>Created on 2019-04-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

```r
# md5 on disk thinks they're different
saveRDS(big_iris, "big_iris.rds", compress = FALSE)
saveRDS(big_iris2, "big_iris2.rds", compress = FALSE)
system("md5sum big_iris.rds")
#> b8905c3310958e4ddc3722326e40eca3 *big_iris.rds
system("md5sum big_iris2.rds")
#> 5c48f7cb1f21fdeb52937d1f3d30ffdf *big_iris2.rds
file.remove("big_iris.rds")
file.remove("big_iris2.rds")
```

Many thanks for this great and useful package. cc @wlandau @richfitz

eddelbuettel:
Insert shrug emoji -- we rely on proper serialization of the object by R.  That may affect `fastdigest` the same way.

Try to see if `serialize()` issues two identical or different objects.

kendonB:
```
> big_iris_ser = serialize(big_iris, NULL)
> big_iris2_ser = serialize(big_iris2, NULL)
> identical(big_iris_ser, big_iris2_ser)
[1] FALSE
```

eddelbuettel:
Interesting.  That looks like work for me then.  

[...]

Issue #33, opened 2016-03-10
WISH: Streamed/incremental digestion?


HenrikBengtsson:
Would it be possible to make a version of `digest()` to incrementally calculate the checksum.  Conceptually something like:

``` r
x <- 1:10
digester <- new_digest(algo="md5")
digester <- add_to_digest(digester, object=x[1:5])
digester <- add_to_digest(digester, object=x[6:10])
checksum <- close_digest(digester)
stopifnot(identical(checksum, digest(x, algo="md5"))
```
# Background

I've got large (~5 GiB) data files that contains ~20-100 genomic sequences.   My goal is to generate a checksum for each individual sequence - so not the whole file.  I know the start and stop file positions of each sequence so I can read them in individually using a file connection (via random access).  What complicates my genomic files is that it contains auxiliary newlines (CR, LF, or CRLF) that are (i) OS dependent and (ii) that may occur at say every 50:th or 80:th character ("different line wrap widths depending on source).  I want my checksum to ignore these newlines.

Yes, I could open a connection, skip to the start of the sequence of interest, read all of the sequence into memory, drop all newlines, and call `digest(clean_seq)`.  However, ideally I'd like to be able to do this with constant-memory constraints.  Hence my wish.

eddelbuettel:
Well, ex ante a lot of things are possible.

Contributors such as yourself have pushed digest beyond what it was originally designed to do. We could keep expanding with clean PRs that do not impact existing interfaces.   

From what you describe, I'd think (very naively) that maybe one needs an intermediating layer watching the stream and 'end of segment' markers before submitting the now-determined segment.  But there may be better ways...

vsimko:
Hi there, the following code covers your requirement of constant memory usage.
Moreover, it also runs parsing the 5GB file and computing the digest in parallel.
Hopefully it helps, although there is certainly space for improvements. (e.g. I'm not sure if this works on Windows)

```r
library(parallel)
library(digest)

x <- as.character(1:10) # input from your example

fname <- "mystream.fifo" # choose name for your named pipe
close(fifo(fname, "w")) # creates your pipe if does not exists

producer <- mcparallel({
    mystream <- file(fname, "w")
    writeLines(x[1:5], mystream)
    writeLines(x[6:10], mystream)
    close(mystream) # sends signal to the consumer (digester)
})

digester <- mcparallel({
    digest(fname, file = TRUE, algo = "md5") # just reads the stream till signalled
})

# runs both processes in parallel
[...]

PR #98, opened 2019-04-22
adds spookyhash

kendonB:
Address #97 and #33.

Still need to update docs and changelog


ChangeLog
```diff
@@ -1,7 +1,3 @@
-2018-12-23  Dirk Eddelbuettel  <edd@debian.org>
-
-	* NAMESPACE: Add .registration=TRUE to useDynLib()
-
```
eddelbuettel: Err. no.

Never *ever* remove from a changelog.

Ever.

NAMESPACE
```diff
@@ -1,5 +1,5 @@
 ## package has a dynamic library
-useDynLib(digest, digest_impl=digest, digest2int_impl=digest2int, AESinit, AESencryptECB, AESdecryptECB, .registration=TRUE)
+useDynLib(digest, digest_impl=digest, digest2int_impl=digest2int, AESinit, AESencryptECB, AESdecryptECB, R_spookydigest)
```
eddelbuettel: Please revert removal of registration.

R/digest.R
```diff
@@ -19,8 +19,8 @@
 ##  along with digest.  If not, see <http://www.gnu.org/licenses/>.
 
 
-digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
-                                  "xxhash32", "xxhash64", "murmur32"),
+digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha256_new", "sha512",
```
eddelbuettel: Please remove addition of sha256_new in the middle.  New ones go to the end.  Do we need `sha256_new` ?

R/digest.R
```diff
@@ -116,13 +129,18 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
...
-                 as.integer(raw),
-                 as.integer(seed))
+    if(algo != "spookyhash"){
+        val <- .Call(digest_impl,
+                     object,
+                     as.integer(algoint),
+                     as.integer(length),
+                     as.integer(skip),
+                     as.integer(raw),
+                     as.integer(seed))
```
eddelbuettel: This looks ... odd.  If 'spookyhash' is selected we use plain `digest_impl` ?

R/digest.R
```diff
@@ -116,13 +129,18 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
...
+    } else {
+        val <- paste(.Call(R_spookydigest, object, NULL, PACKAGE="digest"),
+                     collapse="")
```
eddelbuettel: Why as an overall fallback instead of its own selection?

src/init.c
```diff
@@ -21,18 +21,22 @@
...
 static const R_CallMethodDef callMethods [] = {
     { "PMurHash32", (DL_FUNC) &PMurHash32, 3 },
+    { "R_spookydigest", (DL_FUNC) &R_spookydigest, 2 },
     { NULL, NULL, 0 }
 };
 
 void R_init_digest(DllInfo *info) {
     R_RegisterCCallable("digest", "PMurHash32", (DL_FUNC) &PMurHash32);
-
+    R_RegisterCCallable("digest", "R_spookydigest", (DL_FUNC) &R_spookydigest);
```
eddelbuettel: I know I sent these changes to you. It is not clear we need this.  I do think we had external uses or at least an illustration for `PRMurHash32`.  I think if we kept it I'd want to change the name of `R_spookydigest` to something more standard.  Maybe just `spookydigest_impl` would do.

eddelbuettel:
This PR was a little large and rushed -- it conflated 
- the `R_xlen_t` change (now in va #99)
- addition of spookyhash (which we could still do)
- addition of streaming processing (how can we make sure this isn't tied to spookyhash, would it be alternative like file-based processing became etc pp)
- the addition to registration we actually do not need.

So shall we close this and work on other smaller and more well-defined pieces?

eddelbuettel:
Do we know which, if any, of the existing algos support incremental operations aka 'streaming' ?

PR #100, opened 2019-04-23
add spooky

kendonB:
I believe this addresses your concerns from the last try.

Compared to fastdigest:
 - Removed reference serializer option. This isn't necessary but could be added later.
 - Changed seeds to 0 (seems to be consistent with other implementations). The python package `spooky`, for example, has seed as an argument but the default is `0`. I also removed the `dummy` variable which didn't look like it was doing anything. See:

https://github.com/gmbecker/fastdigest/blob/0c45e8cb748418149cf60a8c9d2b5ee4a1799142/src/serialize2.cpp#L47-L52

and compare:

https://github.com/kendonB/digest/blob/6c0b9764f548207a1b3f660497d82e62f23c71c6/src/spooky_serialize.cpp#L50-L52

 - I haven't been able to match any other spooky hashes (except `fastdigest`) (see https://github.com/gmbecker/fastdigest/issues/3) but I also haven't been able to find a streaming example. May have to add a skip argument?

Still need to add tests but, as I said, I don't have any reference hashes to go by yet.


eddelbuettel:
> Still need to add tests but, as I said, I don't have any reference hashes to go by yet.

Can you add a test or two based on `fastdigest`?

kendonB:
> Can you add a test or two based on `fastdigest`
They use the arbitrary(?): `uint64 seed1 = 100000, seed2 = 9872143234` whereas we currently use 0 for both, which is consistent with other implementations.

eddelbuettel:
That should not matter.

All other tests replicate "published" behaviour (if such reference output exists). 

If we can't do that here (why? maybe the header is different/needs to skipped/maybe streaming gets int the way) then we should 

1. at least make sure fastdigest creates the same and/or 
1. add an option to turn off streaming/headers/.... to ensure we have spookyhash correctly setup.

That's the added confidence tests should give us, besides ensuring continuity through time.

Or am I off the mark here?

kendonB:
Changing the seeds changes the output, of course. Hence fastdigest has different output to the current PR as _I_ changed the seeds to match those used in the spooky python package. _But_, the output also doesn't match the Python package for some reason (see https://github.com/gmbecker/fastdigest/issues/3). We can change the seeds to match fastdigest and then the hashes will match, of course, because that's where I got all the code in the first place.

kendonB:
Would you rather just change the seeds to match fastdigest?

eddelbuettel:
Can the seeding be an argument?  Can we instrument fastdigest for our purposes to have a baseline result to compare against?

kendonB:
We can make seed an argument easily, yes. fastdigest doesn't have seed as an argument but a fork with a one line change could be a good option to have something external. I still need help (or time) to work out why Python spooky is different.

eddelbuettel:
I appreciate your patience in this matter.

kendonB:
FYI, I am going to wait a few days to see if the `fastdigest` author responds before continuing with this.

kendonB:
OK I can match python spooky now if I skip the 30 bytes of header that come with the serialized object. The standard behavior in `digest.c` is to `skip` only 14 bytes. Why is that?

eddelbuettel:
30 bytes versus 14 bytes.  That is ... strange.  While there have been changes to the serialization format over the years this would surprise me.  Have to think about that.

Can you document it in a test file where you compare against a reference string, and maybe show the few lines of Python you used to get the reference string?

eddelbuettel:
Seen this?  https://burtleburtle.net/bob/hash/spooky.html  We can maybe use those tests.  Term is almost over and I should have a little more time coming.  Really appreciate your patience and perseverance here.

kendonB:
Those tests look like they just test ``Spooky::Hash32`` whereas we're always computing 128 bit hashes. 

It looks like the remaining serialization "headers" are associated with object types. So the first 14 bytes must be associated other things like the R version that you wouldn't expect to remain stable over time?

```r
## Includes code that counts bytes as they get processed.
## Comments and alignment with comments added manually
> digest::digest(list(as.raw(1L),as.raw(1L)), algo = "spookyhash")
##first 14##   ##I'm a list##   ##I'm a raw## ##01##  ##I'm a raw## ##01##  
     2 4 4 4              4 4             4 4      1            4 4      1 
[1] "474852d218f1c4e4d363b4154792cccb"
> digest::digest(as.raw(1L), algo = "spookyhash")
##first 14##   ##I'm a raw## ##01##  
     2 4 4 4             4 4      1 
[1] "6979cadb5f772936df55befd698f712d"
```

eddelbuettel:
Look eg at the ChangeLog entries for 2007-03-08 and -09.  I would have to look at R's old NEWS files to see when they made changes but I believe I had to update tests then.

As for what is in the header, I really don't know off the top of my head. and I can't dig now (at work).  If it is pressing for you ask on r-devel.  Someone may answer...

kendonB:
travis fails but "works for me" on Windows and CentOS.

length = number of bytes processed or skipped during this call to `OutBytes`.
skipped = number of bytes skipped before this call to `OutBytes`.
to_skip = total number of bytes that we want to skip.
 
The first object (an empty string) gets processed like this on Windows and my CentOS:
```
length: 2
skipped: 0
to_skip: 30
length: 4
skipped: 2
to_skip: 30
length: 4
skipped: 6
to_skip: 30
length: 4
skipped: 10
to_skip: 30
length: 4
skipped: 14
to_skip: 30
length: 4
skipped: 18
to_skip: 30
length: 4
skipped: 22
to_skip: 30
length: 4
skipped: 26
to_skip: 30
processed this Update(): 0
```

And on Travis:
```
length: 2
skipped: 0
to_skip: 30
length: 4
skipped: 2
to_skip: 30
length: 4
skipped: 6
to_skip: 30
length: 4
skipped: 10
to_skip: 30
length: 4
skipped: 14
to_skip: 30
length: 5
skipped: 18
to_skip: 30
length: 4
skipped: 23
to_skip: 30
length: 4
skipped: 27
to_skip: 30
```
So the serialization header associated with this `character` looks like it has 1 extra byte up front compared to Windows and my CentOS. Recall that we think the first 14 bytes are associated with the R version and only these 14 get skipped in regular calls to digest. What's different about Travis?

kendonB:
I will come back to this in a few days, FYI.

eddelbuettel:
> Still need to work out how to add fastdigest to travis for testing.

Adding `./run.sh install_r fastdigest` should work. `./run.sh` is my variant of the Travis script, it already gets sources.  While `fastdigest` does not exist pre-built (so no `r-cran-fastdigest` for us), it it has not other dependencies so it just installs fine.  

Then again, as I mentioned, I might structure the tests in such a way that we use pre-computed output by `fastdigest` without relying on the package being installed.

eddelbuettel:
You may also need the standard 

```r
if (requireNamespace("fastdigest", quietly=TRUE)) { 
     library(fastdigest) 
      ... 
```

Then again, why not keep it simple and test how we test some of the other digests: against known reference output you just keep as a character string?   Ultimately we probably do not want to depend on   or suggest fastdigest anyway,

kendonB:
Are you happy to just not test against `fastdigest` at all and only test against python spooky? Since `fastdigest` doesn't skip the serialization header and always uses the default serializer, it's output changes from R version to R version.

eddelbuettel:
I had not realized we could not neutralize `fastdigest` against R version changes.  Definitely nicer that it passes now -- will take a closer look but it may take me a few days.

eddelbuettel:
Ok, as you're making more change, we can restore the `.travis.yml` as well. [And you just did while I typed this,  thank you! ]  And it would be nice to just document the python call -- maybe even just in code comments?

The changes to `digest.R` is very clean -- well done.

eddelbuettel:
I'll probably squash merge this tomorrow or Friday. I'll then deal with some small stuff (like updating `digest.Rout.save` with your new output,  adding a bunch of `#nocov` markers etc) and we should have a shiny new release soon thereafter.

PR #102, opened 2019-05-04
Feature/pr100polish

eddelbuettel:
@kendonB Minor postprocessing here.  Take a look at some of the `#nocov` tags and if you can think of ways we _could_ get test coverage there it would of course be preferred to merely tagging.  

I also added a ChangeLog edit for your changes, swushing it all into one April 30 entry.  If you want to expand ...

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -3,3 +3,4 @@
 ^\.Rproj\.user$
 ^\.codecov\.yml$
 ^.*\.tar\.gz$
+^local
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,16 @@
+2019-05-04  Dirk Eddelbuettel  <edd@debian.org>
+
+	* R/digest.R: Add some #nocov tags
+	* src/SpookyV2.h: Idem
+	* src/SpookyV2.cpp: Idem
+	* src/spooky_serialize.cpp: Imde,
+
+2019-05-03  Dirk Eddelbuettel  <edd@debian.org>
+
+	* tests/digestTest.Rout.save: Updated reference output
+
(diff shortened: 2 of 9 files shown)


eddelbuettel:
Also file `src/spooky_serialize.cpp` probably needs a copyright header with attribution to Gabe (and you).

kendonB:
Are you happy not testing the error calls in `digest.R`? Those are, of course, easy to implement.

kendonB:
I believe that we also never use many of the functions in the spooky source that you nocov'd. Would it be better to delete them?

kendonB:
> copyright header with attribution to Gabe

I'm not sure how to do this correctly and would want to be careful with it. Help?

eddelbuettel:
> Are you happy not testing the error calls in digest.R? Those are, of course, easy to implement.
Not exactly sure what you are suggesting, but if it is in answer to my earlier "could we maybe add more tests" then yes :)

> I believe that we also never use many of the functions in the spooky source that you nocov'd. Would it be better to delete them?
Yes, it is a balance between keeping the original source minimally modified to make it clear(er) it is verbatim and genuine ... and making sure we only get what we need.  I think in some other files I had added `#if 0` and later replaced it with `#nocov`.  It more or less ends up meaning the same...

eddelbuettel:
Re the copyright header look maybe at other files and do something like what the `digest.c` file has but credit where it came from.  I eg did that [here in RApiSerialize](https://github.com/eddelbuettel/rapiserialize/blob/master/src/serialize.cpp) for a file that uses code I borrowed from a package from friends who themselves had taken code from R Core in base R.  Just a thought -- we can sketch it here first.   Just want to make sure Gabe and/or R Core gets credit as they should.  I am presuming here you didn't write that file from scratch.  If you did, say so :)

PR #103, opened 2019-05-04
updates to spooky

kendonB:
Let's see how this goes with the coverage bot

I added a few more tests with the right number of bytes to get hit by the missed bits of the algorithm. I also hashed iris which hit some other bits of the algorithm.

Also added a copyright statement at the top of spooky_serialize.cpp

### DIFF
--- a/src/SpookyV2.cpp
+++ b/src/SpookyV2.cpp
@@ -9,9 +9,11 @@
 //   July 30 2012: I reintroduced the buffer overflow
 //   August 5 2012: SpookyV2: d = should be d += in short hash, and remove extra mix from long hash
 //   Kendon Bell April 30 2019: Added counters to help with skipping first n bytes
+//   Kendon Bell May 04 2019: Deleted original source code that doesn't get used
 
 #include <memory.h>
 #include "SpookyV2.h"
+#include "R.h"
 
 #define ALLOW_UNALIGNED_READS 1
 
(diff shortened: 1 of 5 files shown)


src/SpookyV2.cpp
```diff
@@ -9,9 +9,11 @@
 //   July 30 2012: I reintroduced the buffer overflow
 //   August 5 2012: SpookyV2: d = should be d += in short hash, and remove extra mix from long hash
 //   Kendon Bell April 30 2019: Added counters to help with skipping first n bytes
+//   Kendon Bell May 04 2019: Deleted original source code that doesn't get used
 
 #include <memory.h>
 #include "SpookyV2.h"
+#include "R.h"
```
eddelbuettel: Why ?

kendonB: Mistake - I had it there to use Rprintf to see when the various bits of the algorithm got hit

eddelbuettel: Just send a follow up commit.  If need be I merge it again.  (I am new to the PR into PR merge business ;-)  )

eddelbuettel: Else I just do it in the final cleanup.

kendonB:
do you want a PR for the last tweaks or will you do it? A couple of nocovs and removing the include R.h

eddelbuettel:
We can review.  We probably both a few items.  Got to run an errand now -- so no rush.  Thanks for for #103 : additional tests instead of tags "almost never happens" :)

PR #104, opened 2019-05-07
some more polishing

eddelbuettel:
@kendonB Some more polish for `digest`.  See if you have any comments or suggestions.

And sorry for the delays -- teaching term was ending which required some attention from my end.  I should respond quicker now.  We can probably knock this out in a day or two as the ~build~ bulk of the work is done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ src/*.so
 src/*.dll
 inst/doc
 *.tar.gz
+local
+digest.Rcheck
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2019-05-06  Dirk Eddelbuettel  <edd@debian.org>
+
+	* README.md: Mention spookyhash and Kendon
+	* DESCRIPTION (Description): Mention spookyhash
+
 2019-05-04  Dirk Eddelbuettel  <edd@debian.org>
 
 	* R/digest.R: Add some #nocov tags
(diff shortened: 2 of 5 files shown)


eddelbuettel:
Hm, trying to move closer to release shows quite some trouble

###  On Windows tests fails

```
  > for (i in seq(along.with=spookyInput)) {
  +   spooky <- digest(spookyInput[i], algo = "spookyhash")
  +   stopifnot(identical(spooky, spookyOutput[i]))
  +   cat(spooky, "\n")
  + }
  Error: identical(spooky, spookyOutput[i]) is not TRUE
  Execution halted
```

More [here](https://win-builder.r-project.org/nU6x3XfYX5eL/00check.log) at least until the log gets purged.

Also on Fedora at RHub.

### Rhub Debian ASAN/UBSAN error

```
  Comparing ‘digestTest.Rout’ to ‘digestTest.Rout.save’ ...307,354d306                                                                                                                                             
< SpookyV2.h:107:10: runtime error: load of misaligned address 0x556aac75a5a2 for type 'const uint64', which requires 8 byte alignment                                                                             
< 0x556aac75a5a2: note: pointer points here                                                                                                                                                                        
<  14 40  9a 99 99 99 99 99 15 40  66 66 66 66 66 66 14 40  66 66 66 66 66 66 12 40  66 66 66 66 66 66                                                                                                             
<               ^                                                                                                                                                                                                  
< 
```

PR #106, opened 2019-05-11
adjust some spooky tests

kendonB:
The problem was caused by the extra tests I made trying to increase coverage. I removed the reference output for some as they included the serialization header that changes across systems and skipped the first 30 bytes for hashing iris.


kendonB:
Let me clean this up and go straight from "master". Getting on a 12 hr flight rn so might be a day or so

PR #108, opened 2019-05-17
fix skipping bug for streaming_algos

kendonB:
Tests pass. 4x "OK" reports from rhub.

### DIFF
--- a/R/digest.R
+++ b/R/digest.R
@@ -62,14 +62,16 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                       mode=errormode)
     }
 
-    if (serialize && !file && algo %in% non_streaming_algos) {
-        ## support the 'nosharing' option in pqR's base::serialize()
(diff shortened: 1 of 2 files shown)


kendonB:
The change makes sure the `skip` parameter gets set in `digest.R`

eddelbuettel:
Will need one small update for the `digestTest.Rout.save` file; I'll deal with that later.

Thanks again for your persistence.  Really nice work.

eddelbuettel:
Hi @kendonB as you may have seen we're on CRAN now, and gave you a big shout out of thank you (for which I couldn't CC on twitter as you seem twitter-less, and/or my search-fu is weak).

Looks like we are having bad luck with one platform: fedora-clang.  See https://cloud.r-project.org/web/checks/check_results_digest.html  We can probably get to it via rhub.

kendonB:
Thanks to you as well! I am @kmb56 on twitter

eddelbuettel:
And for what it is worth via `rhub::check(".", platform="fedora-clang-devel")` I _cannot_ replicate the failure.

PR #111, opened 2019-07-09
Update README.md

aadler:
simple edits to grammar and prose


eddelbuettel:
Thanks for catching these!  

Always looks super-embarassing after the fact -- so thanks for your help in cleaning it up.

eddelbuettel:
I had forgotten one push (for release 0.6.20) from my box so I force pushed and then cherry-picked your commit back in.

PR #112, opened 2019-07-10
Noticed wrong placement of a bracket

markvanderloo:
to `ignore` the result of a test (i.e. to not record it in the test results) use `ignore(expect_something)(arguments here)`


eddelbuettel:
Nice catch, thank you!

PR #113, opened 2019-09-17
add a vectorised digest function

mdequeljoe:
This pr proposes a function `set_vdigest()` which allows for:

 - setting a vectorised version of `digest()` that is faster than simply using `base::Vectorize()` (this was raised in #2)
 - better performance for repeated calls

`set_vdigest()` returns a function for a given algo and errormode.

With regards to the second point, `set_vdigest()` has been structured such that the information that doesn't need to be repeated in each call is factored out (e.g. matching the algo via `match.arg()` each time)
 
A couple examples:
```r
stretch_key <- function(d, n) {
    md5 <- set_vdigest()
    for (i in seq_len(n))
        d <- md5(d, serialize = FALSE)
    d
}
stretch_key('abc123', 65e3)

sha1 <- set_vdigest(algo = 'sha1')
sha1(letters)
```

The output will always be a character vector so in this respect it differs from `digest` since the `raw` argument is not relevant in this case. I briefly considered whether to output a list if `raw == TRUE` but decided against this. Also I should note that there are some duplicated code blocks now with `digest`, although this has been reduced to some extent with the addition of a 'utils' file.

### benchmarks
```r

library(digest)

#vectorisation
md5 <- set_vdigest()
digest2 <- base::Vectorize(digest)
x <- rep(letters, 1e3)
rbenchmark::benchmark(
    a <- md5(x, serialize = FALSE),
    b <- digest2(x, serialize = FALSE),
    d <- vapply(x, digest, character(1), serialize = FALSE),
    replications = 5
)
all(a == b)
all(a == d)

#repeated calls
stretch_key <- function(d, n) {
    md5 <- set_vdigest()
    for (i in seq_len(n))
        d <- md5(d, serialize = FALSE)
    d
}

stretch_key2 <- function(d, n) {
    for (i in seq_len(n))
        d <- digest(d, serialize = FALSE)
    d
}
rbenchmark::benchmark(
    a <- stretch_key('abc123', 65e3),
    b <- stretch_key2('abc123', 65e3),
    replications = 10
)
a == b
```

### DIFF
--- a/R/vdigest.R
+++ b/R/vdigest.R
@@ -1,5 +1,5 @@
 
-set_vdigest <- function(algo = c("md5", "sha1", "crc32", "sha256", "sha512",
+getVDigest <- function(algo = c("md5", "sha1", "crc32", "sha256", "sha512",
                                  "xxhash32", "xxhash64", "murmur32", "spookyhash"),
                         errormode=c("stop","warn","silent")){
     algo <- match.arg(algo)
--- a/R/vdigest.R
+++ b/R/vdigest.R
@@ -1,10 +1,18 @@
 
-
 set_vdigest <- function(algo = c("md5", "sha1", "crc32", "sha256", "sha512",
                                  "xxhash32", "xxhash64", "murmur32", "spookyhash"),
                         errormode=c("stop","warn","silent")){
     algo <- match.arg(algo)
     errormode <- match.arg(errormode)
+    algoint <- algo_int(algo)
+    non_streaming_algos <- c("md5", "sha1", "crc32", "sha256", "sha512",
+                             "xxhash32", "xxhash64", "murmur32")
+    if (algo %in% non_streaming_algos)
+        return(non_streaming_digest(algo, errormode, algoint))
+    streaming_digest(algo, errormode, algoint)
+}
+
+non_streaming_digest <- function(algo, errormode, algoint){
     function(object,
              serialize=TRUE,
              file=FALSE,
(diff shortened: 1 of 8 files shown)


eddelbuettel:
Nice idea!!  A function `function()`!  I'll peruse.

eddelbuettel:
Nice work.  I just glanced at it on the commuter train in.  A few suggestions

- would you be open to renaming the function to, say, `getVDigest()` ?  No snake_case in the package yet but one camelCase
- what you have above is nice for a demo, I just polished this a little more -- ok if pushed it to your branch and this PR as  `demo/vectorised.R` with `00Index` ?   [ I think we get away without `Suggests:` on `rbenchmark` but I can double check... ]
- I will take care of the coverage stuff afterwards so not to worry there

mdequeljoe:
Thanks for your quick response. The name change is ok with me and sounds good for the demo.

R/digest.R
```diff
@@ -34,17 +34,6 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
...
-        if (mode == "stop") {
-            stop(txt, obj, call.=FALSE)
-        } else if (mode == "warn") {	    # #nocov
-            warning(txt, obj, call.=FALSE)  # nocov
-            return(invisible(NA))           # nocov
-        } else {
-            return(invisible(NULL))         # nocov
-        }
-    }
-
```
eddelbuettel: Moving this to another file is unrelated, and should really only happen after discussion.  

PRs are simply easier to deal with when they are _minimal_.

R/digest.R
```diff
@@ -75,19 +64,8 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
...
-                ## Exclude serialization header (non-data dependent bytes but R
-                ## version specific).  In ASCII, the header consists of for rows
-                ## ending with a newline ('\n').  We need to skip these.
-                ## The end of 4th row is *typically* within the first 18 bytes
-                skip <- which(object[1:30] == as.raw(10))[4] # nocov
-            } else {
-                skip <- 14
-            }
-            ## Was: skip <- if (ascii) 18 else 14
-        }
```
eddelbuettel: See above re non-minimal changes.

R/digest.R
```diff
@@ -104,30 +82,11 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
...
-                      md5=1,
-                      sha1=2,
-                      crc32=3,
-                      sha256=4,
-                      sha512=5,
-                      xxhash32=6,
-                      xxhash64=7,
-                      murmur32=8,
-                      spookyhash=9)
+    algoint <- algo_int(algo)
```
eddelbuettel: See above re non-minimal changes.

R/digest.R
```diff
@@ -104,30 +82,11 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
...
-            return(.errorhandler("The file does not exist: ", object, mode=errormode)) # nocov
-        }
-        if (!isTRUE(!file.info(object)$isdir)) {
-            return(.errorhandler("The specified pathname is not a file: ", # nocov
-                                 object, mode=errormode))                  # nocov
-        }
-        if (file.access(object, 4)) {
-            return(.errorhandler("The specified file is not readable: ",   # #nocov
-                                 object, mode=errormode))                  # #nocov
-        }
```
eddelbuettel: See above re non-minimal changes.

R/utils.R
```diff
@@ -0,0 +1,52 @@
...
+    }
+    if (!isTRUE(!file.info(object)$isdir)) {
+        return(.errorhandler("The specified pathname is not a file: ", # nocov
+                             object, mode=errormode))                  # nocov
+    }
+    if (file.access(object, 4)) {
+        return(.errorhandler("The specified file is not readable: ",   # #nocov
+                             object, mode=errormode))                  # #nocov
+    }
+}
```
eddelbuettel: As previously stated, did these need a new file trampling over 15+ years of commit history?

R/vdigest.R
```diff
@@ -0,0 +1,156 @@
...
+        return(val)
+    }
+}
+
+serialize_ <- function(object, ...){
+    if (length(object))
+        return(lapply(object, base::serialize, ...))
+    base::serialize(object, ...)
+}
+
```
eddelbuettel: I need to look more closely at this but won't have time today.

inst/tinytest/test_digest.R
```diff
@@ -28,6 +28,21 @@ for (i in seq(along.with=md5Input)) {
...
+expect_identical(digest(character(0)),
+                 md5(character(0)))
+expect_identical(digest(list("abc")),
+                 md5(list(list("abc"))))
+expect_identical(digest(list(NULL)),
+                 md5(list(list(NULL))))
+expect_identical(digest(character(0), serialize = FALSE),
+                 md5(character(0), serialize = FALSE))
+
+
```
eddelbuettel: I really appreciate the added tests.

inst/tinytest/test_digest.R
```diff
@@ -266,15 +312,28 @@ for (alg in c("sha1", "md5", "crc32", "sha256", "sha512",
...
+    )
+
     ## Assert that 'skip' works
     h6 <- digest(xskip, algo=alg, serialize=FALSE)
     h7 <- digest(fname, algo=alg, serialize=FALSE, skip=20, file=TRUE)
     expect_true( identical(h6, h7) )
+    expect_identical(
+      set_vdigest(alg)(xskip, serialize = FALSE),
+      set_vdigest(alg)(fname, serialize = FALSE, skip = 20, file = TRUE)
+    )
```
eddelbuettel: As above, great to see tests!

src/digest.c
```diff
@@ -459,3 +459,25 @@ SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Se
...
+        }
+    } else {
+        for (R_xlen_t i = 0; i < n; i++){
+            d = digest(STRING_ELT(Txt, i), Algo, Length, Skip, Leave_raw, Seed);
+            SET_STRING_ELT(ans, i, STRING_ELT(d, 0));
+        }
+    }
+    UNPROTECT(1);
+    return ans;
+}
```
eddelbuettel: This looks good.

R/digest.R
mdequeljoe: Thanks for your feedback. Point well taken on this. I can remove this commit from the PR

eddelbuettel: I would appreciate it.  I clearly see why you did it, and don't disagree on the merits.  But ... smaller diffs are simply better for PRs.  So if you could make that change I'd appreciate it.

Issue #115, opened 2019-09-19
Refactor digest() similar to vdigest


eddelbuettel:
@mdequeljoe : Now that we have a new digest version [which is on its way to CRAN now] thanks to your vectorised approach with the function getter, I want to get back to what you first proposed in your PR.  I sort-of misunderstand the value of it, but it was also lost a little in that PR. It really did two things, and we re-focused on adding vectorised only.

Now, making `digest()` simpler given that you correctly identifies commonalities in helper functions would indeed be a good idea.  

I see that in `vdigest.R` you quite cleverly rewrote / compacted a few functions.  Can I maybe suggest a two step approach that might preserve more of the history?

First, we change `digest()` in `digest.R` to move the functions out.  The names you have in `vdigest.R` are generally sensible.   Once that is done, we maybe replace those functions in a second step (second PR?) with your rewrites.   Does that sound sane, or am I overcomplicating things?

Thanks again for the previous PR.  You are now also listed as a co-author among the ever growing list of contributors.

mdequeljoe:
Thanks for the quick CRAN update. This approach sounds good to me.

eddelbuettel:
[ CRAN update currently on hold as one package ([optmatch](http://cloud.r-project.org/package=optmatch)) came back with a failed test during reverse depends. I could not replicate that, and responded saying so.  We'll see what a human at CRAN says once they get to it.)

Thanks for the quick reply.  So between us we can probably just start a new branch changing `digest.R` first.  I may not get to it for a few days, but it isn't urgent.  Again thanks for the eagle-eyed approach at extending _and_ simplyfying.

eddelbuettel:
Is there actually a second step we need to do or it this the main refactor?

One thing I was wondering was to maybe bring the `utils.c` file in and append to the end of `digest.c` just because that file has a copyright header etc pp (and I failed to keep track on a per-file basis of everybody who contributed....)

Thoughts?   Anything else you think we should before a next release?

mdequeljoe:
Yes on second thought I think this may be sufficient as it is for now. And good point on the copyright header (for the `.R` files), so I think in this regard it would indeed make sense to append this back.

eddelbuettel:
I can take care of it. Poke me here in a few days if I happen to forget about it (currently traveling and will have to catch up to a few other things too).  I'll then roll up a release and get your fine contribution to CRAN.

eddelbuettel:
[...]

PR #116, opened 2019-09-20
start refactor of digest

mdequeljoe:
This PR picks up on part 1 of #115 by using the common helper functions/`if`-blocks in `digest()` that are also used in `vdigest()`. These functions are put in a 'utils.R' file.


eddelbuettel:
[ Because of the then-pending-at-CRAN release of 0.6.21 I had not committed / finalised my changes to DESCRIPTION and ChangeLog, so historically they were "before" your changes here, effectively they were after -- as I  just committed last night.  So that made it a good candidate for `git merge`.  Which I just did.  That preserves your PR as two commits, but moves dates and changes sha1.  The simpler alternative would have been a squash-merge but this we actually had a conflict -- my bad as I asked you to update ChangeLog, and also updated ChangeLog myself.   But it looks like it all worked out. ]

Issue #118, opened 2019-09-30
CFB cipher implementation 


isuruceanu:
Is there any plan to support [Cipher Feedback (CFB)](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Cipher_Block_Chaining_(CBC))?

eddelbuettel:
Not yet.  We could have you an implementation with a compatible / suitable license, and maybe also test functions and/or input and output.  We have extended the package in the past to other ciphers via suitable PRs.  Do you want to work on this?

I only glanced at the linkL what does CFB do for you?  Do you use it another context? Or speed?

isuruceanu:
We have an app written in golang which collect some data and encrypt it using aes.CFB cipher. We want to analyze and use it for some ML, because of luck good ML packages in golang we are looking to use python or R. Python has an implementation of CFB cipher which is fine, just looking for R package too.

Will try to implement it, although haven't done any c/c++ for along time :)

eddelbuettel:
Sounds good.  It _should_ be straightforward as we (generally) just take a `char*` in (with the already serialized R object) and return one with the digest.  Poke us here if you have questions.  Several of the existing digests should providing working examples.

PR #119, opened 2019-10-01
feat: implement CFB cipher mode

isuruceanu:
Support [Cipher Feedback (CFB) mode](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Cipher_Feedback_(CFB)) 

CFB uses a block cipher as a component of a random number generator. In CFB mode, the previous ciphertext block is encrypted and the output is XORed (see XOR) with the current plaintext block to create the current ciphertext block. The XOR operation conceals plaintext patterns. Plaintext cannot be directly worked on unless there is retrieval of blocks from either the beginning or end of the ciphertext.

fix: #118

### DIFF
--- a/R/AES.R
+++ b/R/AES.R
@@ -1,13 +1,13 @@
 # This is loosely modelled on the AES code from the Python Crypto package.
 
-# Currently only ECB, CBC and CTR modes are supported...
+# Currently only ECB, CBC, CFB and CTR modes are supported...
 
 modes <- c("ECB", "CBC", "CFB", "PGP", "OFB", "CTR", "OPENPGP")
 
-AES <- function(key, mode=c("ECB", "CBC", "CTR"), IV=NULL) {
+AES <- function(key, mode=c("ECB", "CBC", "CFB", "CTR"), IV=NULL) {
     mode <- match(match.arg(mode), modes)
-    if (!(mode %in% c(1:2, 6)))
-        stop("Only ECB, CBC and CTR mode encryption are supported.")	# #nocov
+    if (!(mode %in% c(1:3, 6)))
+        stop("Only ECB, CBC, CFB and CTR mode encryption are supported.")	# #nocov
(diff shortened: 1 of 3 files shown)


eddelbuettel:
There is something weird going on with this PR -- please see the diff yourself at https://github.com/eddelbuettel/digest/pull/119/files. This makes the comparison of "what changed" rather difficult.  [ Goes off checking. ]   Ok, if one goes to view the diff and then selects settings therein one can again select 'ignore whitespace' and that makes it better.

Looks reasonable, and it appears we supported CFB all along and your PR "only" needed to turn it on, document it, and test it.  Nice.

eddelbuettel:
Do you have a reference for the test input / output?   If so, could you add it?

isuruceanu:
> There is something weird going on with this PR
noted too, will investigate and fix 

> Do you have a reference for the test input / output? If so, could you add it?

Thanks for pointing. Will update tests and man with same examples from [ SP800-38a](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf#page=36)


PS: will be good to squash commits into one once merge the PR

eddelbuettel:
> noted too, will investigate and fix

Sometimes it is as simple as editors used.  

NIPS reference is perfect; we have that for some other tests too.   As for squashing:  a handful of related commits is fine by me.  I often run `git log` with a tree display and (especially as long as it is somewhat orderly without parallel development threads crossing) I find the extra detail informative. Advantage of still somewhat smaller open source projects :)

isuruceanu:
I've checked once again unit test, for cfb128  use same plaintext as for overs tests and results are identical to NIPS 

man AES.Rm same as NISP too.

eddelbuettel:
Thanks, time to merge; I can always do a small cleanup and add eg the NIST reference.

PR #121, opened 2019-10-12
Fix sha1(Inf)

### LINKED ISSUE #120: sha1() fails on Inf
When running `sha1()` on `Inf`, an error occurs.

``` r
library(digest)
sha1(Inf)
#> Warning in num2hex(x, digits = digits, zapsmall = zapsmall): NAs introduced
#> by coercion to integer range
#> Error in if (any(zero)) {: missing value where TRUE/FALSE needed
```

<sup>Created on 2019-10-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

billdenney:
Fix #120 

`num2hex(Inf)` now returns `"Inf"`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
-2019-10-12  Bill Denney <wdenney@humanpredictions.com>
+2019-10-13  Bill Denney <wdenney@humanpredictions.com>
 
-  * R/sha1.R: Fix issue with num2hex() and input of Inf
-  * inst/tinytest/test_sha1.R: Add Inf to the tests
+	* R/sha1.R: Fix issue with num2hex() and input of Inf and -Inf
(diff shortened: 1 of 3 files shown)


billdenney:
@ThierryO, can you please review, too?

billdenney:
I found an issue that `-Inf` was considered the same as `Inf`.  It is now fixed.

eddelbuettel:
Good catch, and smart of me to be out and about at the Chicago Marathon all day.

Guess I'll merge shortly.

billdenney:
That Chicago Marathon day was some of the best defensive programming that I've seen!  :)

eddelbuettel:
Master plan didn't work all that well -- but I just hand-edited this directly in the other branch.  

Can you take a peek there to double-check?

PR #124, opened 2019-10-15
Support UTF-8 encoded file paths on Windows

### LINKED ISSUE #89: Cannot open input file when filename contains umlaut and file = TRUE
Consider two files which has the same contents, only differing by filename:

```
> dir("data")
[1] "Gavleborg.txt" "Gävleborg.txt"
```

digest::digest works on the file without the "ä".
```
> digest(file.path("data", "Gavleborg.txt"), algo = "md5", file = TRUE)
[1] "85d6d41527468fd765392c6284661e62"
```

But throws an error when used on the file with the "ä", even when replacing ä with \u00E4. Note that the error message contains the correct character, but not the line after the colon.
```
> digest(file.path("data", "Gävleborg.txt"), file = TRUE)
Error in digest(file.path("data", "Gävleborg.txt"), file = TRUE) : 
  Cannot open input file: data/GÃ¤vleborg.txt

> digest(file.path("data", "G\u00E4vleborg.txt"), file = TRUE)
Error in digest(file.path("data", "Gävleborg.txt"), file = TRUE) : 
  Cannot open input file: data/GÃ¤vleborg.txt
```

However, it works when the "file"-argument is removed
```
> digest("data/Gävleborg.txt")
[1] "a8c6abd369f1b7e60c10fd35df55b123"
```

The function is passed with file = TRUE when I deploy a Shiny app to shinyapps.io, so I'm not sure whether this should be considered a bug within the shiny deployment context instead.

jimhester:
This assumes the file paths will be UTF-8 encoded, they most likely
will be with recent versions of R. The paths are converted to UTF-16 by
`MultiByteToWideChar()` before being passed to `_wfopen()`.

On other systems we call `fopen()` directly, as the code did before.

I also pulled out the individual calls to `fopen()` and put them above
the switch to reduce some repetition in the code.

There is a also a simple test which demonstrates the issue.

Fixes #89

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,7 @@
-2019-10-13  Bill Denney <wdenney@humanpredictions.com>
+2019-10-13  Jim Hester <james.f.hester@gmail.com>
+
 	* src/digest.c: Add support for UTF-8 file paths on Windows
+	* R/digest.R: Idem
 	* inst/tinytest/test_encoding.R: Add tests for UTF-8 file paths.
 
 2019-10-13  Bill Denney <wdenney@humanpredictions.com>
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,8 @@
 2019-10-13  Bill Denney <wdenney@humanpredictions.com>
+	* src/digest.c: Add support for UTF-8 file paths on Windows
+	* inst/tinytest/test_encoding.R: Add tests for UTF-8 file paths.
+
+2019-10-13  Bill Denney <wdenney@humanpredictions.com>
 
 	* R/sha1.R: Add sha1.formula()
 	* NAMESPACE: Idem
--- a/src/digest.c
+++ b/src/digest.c
@@ -37,12 +37,45 @@
 #include "xxhash.h"
 #include "pmurhash.h"
 
+#ifdef _WIN32
+#include <Windows.h>
+#endif
+
 unsigned long ZEXPORT digest_crc32(unsigned long crc,
                                    const unsigned char FAR *buf,
                                    unsigned len);
 
 static const char *sha2_hex_digits = "0123456789abcdef";
 
+#ifdef _WIN32
+FILE* my_open(const char* txt) {
+    FILE* out;
+    wchar_t* buf;
+    size_t len = MultiByteToWideChar(CP_UTF8, 0, txt, -1, NULL, 0);
+    if (len <= 0) {
+        error("Cannot convert file to Unicode: %s", txt);
+    }
+    buf = (wchar_t*) R_alloc(len, sizeof(wchar_t));
+    if (buf == NULL) {
+        error("Could not allocate buffer of size: %ll", len);
+    }
+
+    MultiByteToWideChar(CP_UTF8, 0, txt, -1, buf, len);
+
+    if (!(out = _wfopen(buf, L"rb"))) {
+        error("Cannot open input file: %s", txt); /* already covered at R level too */ /* #nocov */
+    }
+    return out;
+}
+#else
+FILE* my_open(const char* txt) {
+    FILE* out;
+    if (!(out = fopen(txt, "rb"))) {
+        error("Cannot open input file: %s", txt); /* already covered at R level too */ /* #nocov */
+    }
+    return out;
+}
+#endif
 
 SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Seed) {
     size_t BUF_SIZE = 1024;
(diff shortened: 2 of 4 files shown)


ChangeLog
```diff
@@ -1,3 +1,7 @@
+2019-10-13  Bill Denney <wdenney@humanpredictions.com>
```
gaborcsardi: I guess you wanted to update this. :)

eddelbuettel:
Hadn't noticed the redundancy with the multiple assignments to `fb`, that alone is worth fixing.

I know little about the Windows issue.  Does this mean that we will to similarly patch every single package calling `fopen()` somewhere?  Ouch.  Too bad there is no common API for this.

gaborcsardi:
There is `RC_fopen()` for this, but it is not part of the public API it seems.

jimhester:
Yeah, R has had [RC_fopen()](https://github.com/wch/r-source/blob/e732fe187ee0d680da79330347c2c5fb0ce1da11/src/main/sysutils.c#L198-L205) for 10+ years, but it is not in the public API :(

eddelbuettel:
Could we do what I did with RApiDatetime abd RApiSerialize and just "borrow" the GPL'ed code and offer the function at the C level?  Seems better than copying ...

gaborcsardi:
I think we could, but Jim's new function is simpler and just as good, so personally I think it is just easier to copy it over, instead of calling C code across packages. But up to you. And btw. we could also ask R core for some public API.

eddelbuettel:
Right.  And I wasn't volunteering, I have enough packages as it stands, and zero interest in Windows-only hackery.  For what it is worth, I think copying is bad as it will lead to stale disjoint instances.  Oh well.

src/digest.c
```diff
@@ -37,12 +37,45 @@
...
+
+    if (!(out = _wfopen(buf, L"rb"))) {
+        error("Cannot open input file: %s", txt); /* already covered at R level too */ /* #nocov */
+    }
+    return out;
+}
+#else
+FILE* my_open(const char* txt) {
+    FILE* out;
+    if (!(out = fopen(txt, "rb"))) {
```
eddelbuettel: This is the only line that is different.  So how about rewriting this to

```c++
FILE* my_open(const char* txt) {
￼  FILE* out;
#ifdef _WIN32
   wchar_t *buf;
  // ... as a above
   out = _wfopen(buf, L"rb");
#else 
   out = fopen(txt, "rb")
#endif
  if (!out)  {
       error("Cannot open input file: %s", txt);  /* #nocov */
   }
   return out;
}
``` 

which keeps the per-platform code more local?   And come to think about it, I probably take the error check out as well given that we only need it one place (as you righty identified).  

And then, while we're at it, maybe change the not-too-descriptive `my_open` to `open_with_windows_widechar` or something.

src/digest.c
```diff
@@ -67,6 +100,10 @@ SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Se
     } else { /* or a string */
         txt = (char*) STRING_VALUE(Txt);
         nChar = strlen(txt);
+
+        if (algo >= 100) {
+            fp = my_open(txt);
+        }
```
eddelbuettel: We can the same with `fclose(fp)` which also litter every block.

jimhester:
Thanks for reviewing! I did as you suggest in 7cadf27

eddelbuettel:
@jimhester No rest for the wicked, and the usual reminder that the road to hell is in fact paved with good intentions, and failing tests.  This went belly up on (Kurt's fairly standard) clang test box:

https://cloud.r-project.org/web/checks/check_results_digest.html

Any idea why?

gaborcsardi:
Some Unix platforms are also not UTF-8, e.g. one of CRAN's builders. So we should only call `enc2utf8()` on Windows, and also only run the test case on Windows.

You can probably reproduce this on the `debian-clang-devel` R-hub builder, btw.

eddelbuettel:
Good idea, but `preperror`.  Race condition?  
https://builder.r-hub.io/status/digest_0.6.22.tar.gz-5399bd9689d248c69f05c5a6bf62ebc8

eddelbuettel:
Until we have a better idea I added this in the sources

```diff
modified   inst/tinytest/test_encoding.R
@@ -1,3 +1,7 @@
+
+isWindows <- Sys.info()[["sysname"]] == "Windows"
+if (!isWindows) exit_file("Skipping test on non-Windows platform")
+
 d <- tempfile()
 dir.create(d)

```

jimhester:
I think that if in the test plus putting only doing `enc2utf8()` on windows should do it, so at https://github.com/eddelbuettel/digest/blob/4c415fe2b67b760f7cd64767620d561a5a1a0ed6/R/digest.R#L93 also doing

```r
object <- path.expand(object)
if (isWindows) {
  object <- enc2utf8(object)
}

eddelbuettel:
Done in https://github.com/eddelbuettel/digest/commit/caa165d23d690e5c7ca71194b477edeebf24ce77

Issue #126, opened 2019-10-25
Object Attributes Are Ignored with `sha1.list()`


billdenney:
Attributes for list objects are ignored with `sha1.list()`.  The hash for `baz` and `zot` below seem like they should differ.  I haven't audited to check if this would apply more generally, but I assume that any object with differing attributes should have differing hashes.

In a more complex consideration, perhaps some level of attribute checking should be performed across objects so that if, for instance, I added an attribute to a POSIXlt object, the attribute would be considered, even though POSIXlt objects already have attributes as a component of the object itself.  (Hopefully that is clear.)

``` r
library(digest)
baz <- list(A=1)
zot <- baz
attr(zot, "abc") <- 1
sha1(baz)
#> [1] "6c3665e1d49d3cece26ee7ae3ede1b14e8e7d0b5"
sha1(zot)
#> [1] "6c3665e1d49d3cece26ee7ae3ede1b14e8e7d0b5"
```

<sup>Created on 2019-10-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

See also: #125

Tag: @ThierryO, @eddelbuettel

eddelbuettel:
If we decide to change this I should probably run a reverse-depends check. 

Then again, most uses of the package are "probably"  `digest::digest()` rather than `digest::sha1()` but it will be prudent to check.  Which I can do with the machinery I use for Rcpp and friends.

billdenney:
As an intermediate option here, we could add an `attributes=TRUE` default option which could be set to `FALSE` for the old behavior (showing my bias that attributes are part of the object and should be included).

eddelbuettel:
Good point.  And we could feed the option default value with a global `options()` value, and/or possibly phase the change in.  So lots of change managment options _should we need them_.

ThierryO:
A potential issue is that `sha1` itself adds attributes to the object. How should handle those? Especially in cases where the hash is calculcated as a hash of hashes (e.g. the hash of a list is the hash of the vector of hashes of its elements).

ThierryO:
Looking a bit deeper on this, `sha1()` only sets some attributes and does some transformations if needed. Then the transformed object is passed to `digest()`. So if we would implement the attributes, should it be `sha1()` or `digest()` that handles them?

eddelbuettel:
We could experiment with one, see if it works, backtrack if not ...

ThierryO:
I had a better look at the code and must correct myself.
[...]

PR #127, opened 2019-10-29
Sha1 attributes

### LINKED ISSUE #125: Generalization of sha1.default
As I was trying to improve the `sha1` for the PKNCA package, I found that I needed `sha1.sessionInfo()` and `sha1.packageDescription()` functions.  Those would be just be:

```r
sha1.sessionInfo <- digest:::sha1.list
sha1.packageDescription <- digest:::sha1.list
```

That made me think that a generalization of the `sha1.default()` function to check with something like the following would help:

```r
if (is.list(x)) {
  sha1.list(x=x, digits = digits, zapsmall = zapsmall, ..., algo = algo)
} else {
  # The current stop message
}
```

And, the one other change that may be of interest would be that `sha1.list()` pull out the attributes of the list for checking, too.  (Though, that looks like a different issue that I'll open separately.)

https://github.com/eddelbuettel/digest/blob/4c415fe2b67b760f7cd64767620d561a5a1a0ed6/R/sha1.R#L28-L35

Would such a PR be of interest, or would you prefer the specific methods for each class?  (And therefore you would prefer a PR that explicitly names the sessionInfo and packageDescription classes.)

Tag: @ThierryO, @eddelbuettel

ThierryO:
fixes #125 and #126

### DIFF
--- a/inst/tinytest/test_sha1.R
+++ b/inst/tinytest/test_sha1.R
@@ -467,3 +467,5 @@ for (algo in c("md5", "sha1", "crc32", "sha256", "sha512", "xxhash32",
     )
 
 }
+
+expect_true(is.character(sha1(sessionInfo())))
--- a/inst/tinytest/test_sha1.R
+++ b/inst/tinytest/test_sha1.R
@@ -70,6 +70,7 @@ expect_true(
         }
     )
 )
+options(sha1PackageVersion = "0.6.22.1")
 expect_true(
     identical(
         sha1(x.dataframe),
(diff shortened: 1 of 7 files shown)


eddelbuettel:
Looks very clean, thank you!   (I may tweak the docs and version comparison to make 0.6.23 as interim versions "do not really exist" in the on-CRAN and in  the big wide world sense.)

eddelbuettel:
The codecov change is from the is.list check:  

```r
    if (is.list(x)) {
        return(
            sha1.list(x, digits = digits, zapsmall = zapsmall, ..., algo = algo)
        )
    }
```

Can you throw a quick unit test in for a list object ?

eddelbuettel:
Done now in https://github.com/eddelbuettel/digest/pull/127/commits/59ba9fb4268411866759bec71b60faa24d0ebdd0

eddelbuettel:
The test not working reveals that there is of course a `sha1.list()` method so will that code in `sha1.default()` ever work?

ThierryO:
a `sessionInfo()` is a list like object and shoudl be handled by `sha1.default()`

inst/tinytest/test_sha1.R
```diff
@@ -467,3 +468,19 @@ for (algo in c("md5", "sha1", "crc32", "sha256", "sha512", "xxhash32",
     )
 
 }
+
```
ThierryO: unit test for `sha.default()` with `is.list(x) == TRUE`
```suggestion
expect_true(is.character(sha1(sessionInfo())))
```

ThierryO:
I've added the unit test with an interactive rebase to fe2f6f8b.
The documentation update (65d1875d) now states versions 0.6.22 and 0.6.23

eddelbuettel:
I still see the 'four-part' version in the rd file but no worries, I can smooth over that when I roll up the next release (unless I forget).

Thanks again--will merge.

eddelbuettel:
> I still see the 'four-part' version in the rd file 

Must have been a web-page caching artefact. All good.  Thanks again!

billdenney:
Thanks to both of you for your help on these issues!

PR #129, opened 2019-11-04
Add sha1.( support

### LINKED ISSUE #128: Support for the "(" class
Within certain formula, the "(" class is used for parenthetical notation.  I think that it's mainly (or perhaps only) when the parenthesis is the first item in the formula.  Regardless, the fix is pretty simple.  (I learn new things about R all the time; I had assumed that formula were made only of names and calls.)

``` r
library(digest)
sha1(a~(c+d))
#> Error: sha1() has no method for the '(' class
class((a~(c+d))[[3]])
#> [1] "("
"sha1.(" <- digest:::sha1.formula
sha1(a~(c+d))
#> [1] "3975451db3604aa3f8a947af38f29ed7d3930cb3"
```

<sup>Created on 2019-11-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

billdenney:
This fixes #128, but it has an issue with the documentation that appears to relate to the fact that the class name is "(".  I can't figure that one out with several tries for escaping it (nothing, quotes, and backslash all tried).

### DIFF
--- a/R/sha1.R
+++ b/R/sha1.R
@@ -449,3 +449,5 @@ sha1.formula <- function(x, digits = 14L, zapsmall = 7L, ..., algo = "sha1"){
     )
     digest(y, algo = algo)
 }
+
+"sha1.(" <- sha1.formula
--- a/man/sha1.rd
+++ b/man/sha1.rd
@@ -22,6 +22,7 @@
 \alias{sha1.call}
 \alias{sha1.raw}
 \alias{sha1.formula}
+\alias{sha1.(}
 \title{Calculate a SHA1 hash of an object}
 \author{Thierry Onkelinx}
 \usage{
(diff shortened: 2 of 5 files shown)


eddelbuettel:
I am calling this a teachable moment. What is a dispatch on `(`? Do you have an example?  Is this for expressions "on the fly"?

man/sha1.rd
```diff
@@ -90,5 +92,5 @@ with \code{sha1(x, algo = "sha1")}, they will be different starting from digest
 
 Until version 0.6.22, \code{sha1} ignored the attributes of the object for
 some classes. This was fixed in version 0.6.23. Use
-\code{option(sha1PackageVersion = 0.6.23)} to get the old behaviour.
+\code{option(sha1PackageVersion = 0.6.22)} to get the old behaviour.
```
eddelbuettel: Good catch!

R/sha1.R
```diff
@@ -449,3 +449,5 @@ sha1.formula <- function(x, digits = 14L, zapsmall = 7L, ..., algo = "sha1"){
     )
     digest(y, algo = algo)
 }
+
+"sha1.(" <- sha1.formula
```
eddelbuettel: So ... that line is the gist of the PR?  Kewl.

billdenney: Yep, it's pretty simple.

ThierryO: Elegant solution

PR #131, opened 2019-11-07
Add sha1_digest() and sha1_attr_digest for code simplification (closes #132)

### LINKED ISSUE #130: Code Simplification
As I've been looking a lot at the code for `sha1()` over the past few weeks, I realized that there is some duplication there that could be simplified.  I did a programmatic look over the functions in the package, and I found that the following are all identical functions:

* "sha1.call" "sha1.character" "sha1.factor"    "sha1.integer"   "sha1.logical"   "sha1.raw"      
* "sha1.name" "sha1.NULL"

Would you be interested in a PR that combines these together into two functions (named something like `sha1_attr_digest()` and `sha1_digest()`) which would cover the general cases and then have a simple assignment so that all of these classes are covered?  (To make this more concrete, I would rename the `sha1.call()` function to `sha1_attr_digest()` then make an assignment of `sha1.call <- sha1.character <- sha1.factor <- sha1.integer <- sha1.logical <- sha1.raw <- sha1_attr_digest`.  (And similar with `sha1.name()` becoming `sha1_digest()` and assigning `sha1.name <- sha1.NULL <- sha1_digest`.)

The benefit that I see is that there are fewer lines of code (removing repetition), future classes that can follow one of these two paradigms could simply be added by an additional assignment, and if a structural change were made in the future, it would require fewer changes to ensure consistency.

(An underlying goal with this is a "thank you PR" for your support in my recent requests!)

Code to find this was:

```r
library(digest)

all_funs <- unclass(lsf.str(envir = asNamespace("digest"), all.names = TRUE))

remaining_funs <- all_funs
identical_funs <- list()
idx1 <- 1
while (idx1 < length(remaining_funs)) {
  idx2 <- idx1 + 1
  while (idx2 <= length(remaining_funs)) {
    if (identical(x=get(x=remaining_funs[idx1], envir=asNamespace("digest")),
                  y=get(x=remaining_funs[idx2], envir=asNamespace("digest")))) {
      identical_funs[[remaining_funs[idx1]]] <-
        c(identical_funs[[remaining_funs[idx1]]], remaining_funs[idx2])
      remaining_funs <- remaining_funs[-idx2]
    } else {
      idx2 <- idx2 + 1
    }
  }
  idx1 <- idx1 + 1
}
identical_funs
```

### LINKED ISSUE #132: Documentation issue for sha1.(
I didn't make this clear enough when I wrote the sha1.( method.  There is a documentation issue which I could not figure out on how to add the `sha1.(` method to the .Rd file.  When checking the code, I get the following error:

```
> checking Rd \usage sections ... WARNING
  Bad \usage lines found in documentation object 'sha1':
    <unescaped bksl>method{sha1}{"("}(x, digits = 14, zapsmall = 7, ..., algo = "sha1")
  
  Functions with \usage entries need to have the appropriate \alias
  entries, and all their arguments documented.
  The \usage entries must correspond to syntactically valid R code.
  See chapter 'Writing R documentation files' in the 'Writing R
  Extensions' manual.
```

It relates to this line of documentation:

https://github.com/eddelbuettel/digest/blob/f0b6fa91d0f2efe4223a677ad41018885b92a248/man/sha1.rd#L51

I tried several variants of escaping the parenthesis (`"("`, `\(`, and `\\(`) without success.

billdenney:
Fix #130 
Fix #132

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,10 +1,16 @@
+2019-11-07  Bill Denney <wdenney@humanpredictions.com>
+
+  * NAMESPACE: add sha1_digest() and sha1_attr_digest() Functions
+  * R/sha1.R: Idem
+  * man/sha1.Rd: Idem
+  * Also, renamed sha1.rd to sha1.Rd
+
 2019-11-03  Bill Denney <wdenney@predictions.com>
 
 	* R/sha1.R: Add support for the "(" class which is used in some formula
 	* NAMESPACE: Idem
 	* man/sha1.Rd: Idem; also corrected typo for old version behavior
 	* inst.tinytest/test_sha1.R: Add tests for the "(" class
-	* 
 
 2019-10-27  Thierry Onkelinx  <thierry.onkelinx@inbo.be>
 
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -9,6 +9,8 @@ export(AES,
        digest2int,
        getVDigest,
        sha1,
+       sha1_attr_digest,
+       sha1_digest,
        hmac,
        makeRaw)
 
--- a/R/sha1.R
+++ b/R/sha1.R
@@ -39,39 +39,6 @@ sha1.default <- function(x, digits = 14L, zapsmall = 7L, ..., algo = "sha1") {
     )								# #nocov end
 }
 
-sha1.integer <- function(x, digits = 14L, zapsmall = 7L, ..., algo = "sha1") {
-    attr(x, "digest::sha1") <- attr_sha1(
-        x = x, digits = digits, zapsmall = zapsmall, algo = algo, ...
-    )
-    digest(x, algo = algo)
-}
-
-sha1.character <- function(x, digits = 14L, zapsmall = 7L, ..., algo = "sha1") {
-    attr(x, "digest::sha1") <- attr_sha1(
-        x = x, digits = digits, zapsmall = zapsmall, algo = algo, ...
-    )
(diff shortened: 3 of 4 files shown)


eddelbuettel:
@billdenney I did two quick checks on the short train commute in:

1) I do not think we need the two lines in `NAMESPACE` as I think those are internal functions and not exported
1) I just remove the `sha1.(` line the in the Rd to no ill effect :)  and no warning

Thoughts?

billdenney:
@eddelbuettel 

1. I was thinking that those functions would be recommended as simple functions for other packages to use as their generic function when they want an `sha1()` method, so I would prefer to keep them.  If we do not export them, then we will need to revise the documentation accordingly.  (My plan for them was to use `sha1_attr_digest()` in my PKNCA package for the objects where I want hashes.)
2. I debated suggesting that.  :)  I really doubt that anyone will need `sha1.(` exported, but I am also persistently surprised by such things.  I've asked on r-package-devel, and if it gets a quick answer, I think that the best option is to fix it.  If a slow answer or if it's a bug in R, I think drop it.

Do those sound reasonable?

billdenney:
Duncan had a quick reply that fixed `sha1.(` documentation.  (Fixes #132)

eddelbuettel:
Saw the thread with Duncan on r-package-devel, very good.

Your answer on whether or not to export these is fine too: you appear to have a use, so let's export (as you would not be allowed to access via `:::`).

Issue #133, opened 2019-11-11
Improve sha1 efficiency


billdenney:
Probably a question for @ThierryO.

I needed to generate a hash on a data.frame with about 1M rows and about 5 numeric columns (so, about 5M total numeric values), and it took a bit of time (many seconds).  So, I was looking for where I could help make things faster, and the answer appears to come in `num2hex()`.

At a first look, most of the time is spent in various forms of string manipulation, and as I dove down the rabbit hole of trying to make those string manipulations faster, I found the C++ function `frexp` which extracts the mantissa and exponent.  I wonder if we could substitute most of the more time-intensive string manipulation in `num2hex()` with a simpler C++ call to `frexp` (and then also do the hex generation there, too, using the integer form)?

eddelbuettel:
From a quick glance at Google this seems to be a well-travelled area. Not sure we can pick up much.

One link:  https://crypto.stackexchange.com/questions/60346/fastest-way-to-compute-iterates-of-sha1

billdenney:
Thanks for that research.  While my goal is general, at least in this first case for num2hex, I’m working on the steps before the actual hash calculation.  In my quick profiling of larger numeric vectors, over 90% of the time is spent in sprintf and gsub to extract the negative sign, mantissa, and exponent.  I’m wanting to make that part more efficient as the first step.

billdenney:
A second way to consider this issue is a general request to look for efficiency improvements.  The package works well, and it appears to be generally stable and feature-complete, so trying for more efficiency seems reasonable to me.  (Selfishly says the guy who is using it for operations taking many seconds to minutes...)

eddelbuettel:
Poor man's efficiency is just throwing it at a bazillions cores :)

billdenney:
Ahh, to be so poor!  :)

billdenney:
I think that the `num2hex()` update is what I have a reasonable chance of finding.  I'm going to close this.

PR #135, opened 2019-11-11
Speed up num2hex by removing some string operations

billdenney:
This is at least part of #133.

Results should be identical, and it should be ~40% faster:

```r
> microbenchmark::microbenchmark(
+   old_num2hex(x),
+   num2hex(x)
+ )
Unit: milliseconds
           expr      min       lq     mean   median       uq      max neval
 old_num2hex(x) 122.0020 138.5887 165.2797 154.6198 170.3841 297.2866   100
     num2hex(x)  77.0894  89.3238 103.2708  97.0571 113.1322 173.4948   100
```

### DIFF
--- a/R/sha1.R
+++ b/R/sha1.R
@@ -398,30 +398,35 @@ num2hex <- function(x, digits = 14L, zapsmall = 7L){
     output <- rep(NA_character_, length(x))
     output[x.inf & x > 0] <- "Inf"
     output[x.inf & x < 0] <- "-Inf"
-    x.finite <- !x.na & !x.inf
-    x.hex <- sprintf("%a", x[x.finite])
-    exponent <- as.integer(gsub("^.*p", "", x.hex))
     # detect "small" numbers
-    zapsmall.hex <- floor(log2(10 ^ -zapsmall))
-    zero <- x.hex == sprintf("%a", 0) | exponent <= zapsmall.hex
-    if (any(zero)) {
-        output[x.finite][zero] <- "0"
-        if (all(zero)) {
-            return(output)
-        }
+    x.zero <- !x.na & !x.inf & abs(x) <= (2^floor(log2(10 ^ -zapsmall)))
+    if (any(x.zero)) {
+        output[x.zero] <- "0"
     }
+    x.finite <- !(x.na | x.inf | x.zero)
+    # The calculations for non-na, non-inf, non-zero values are computationally
+    # more intense.  Don't do them unless necessary.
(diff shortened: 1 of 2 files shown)


billdenney:
@ThierryO, I would definitely appreciate your review here as you understand the complexities of num2hex very well (and based on other conversations here, probably best).

R/sha1.R
```diff
@@ -398,30 +398,35 @@ num2hex <- function(x, digits = 14L, zapsmall = 7L){
...
-    zapsmall.hex <- floor(log2(10 ^ -zapsmall))
-    zero <- x.hex == sprintf("%a", 0) | exponent <= zapsmall.hex
-    if (any(zero)) {
-        output[x.finite][zero] <- "0"
-        if (all(zero)) {
-            return(output)
-        }
-    }
+    x.zero <- !x.na & !x.inf & abs(x) <= (2^floor(log2(10 ^ -zapsmall)))
+    if (any(x.zero)) {
```
ThierryO: I think we can even drop this `ìf` statement as we don't have it in lines 399 and 400. We just need to make sure that is works as expected when `x.zero` contains only `FALSE` values.

R/sha1.R
```diff
@@ -398,30 +398,35 @@ num2hex <- function(x, digits = 14L, zapsmall = 7L){
...
-    # remove potential trailing zero's
-    mantissa <- gsub(mantissa, pattern = "0*$", replacement = "")
-    negative <- ifelse(grepl(x.hex[!zero], pattern = "^-"), "-", "")
-    output[x.finite][!zero] <- paste0(negative, mantissa, " ", exponent[!zero])
+    # select mantissa (starting format is 0x1.[0-9a-f]+p[+-][0-9]+), remove the
+    # beginning through the decimal place, including the fact that exact powers
+    # of two will not have a decimal place.
+    mantissa <- x.hex
+    # Remove the beginning through the decimal place (if it exists).
+    mask_decimal <- startsWith(mantissa, "0x1.")
```
ThierryO: This makes line 421 obsolete

```suggestion
    mask_decimal <- startsWith(x.hex, "0x1.")
```

R/sha1.R
billdenney: I just verified with

```r
> output <- "1"
> x.zero <- FALSE
> output[x.zero] <- "0"
```

which worked as expected.

R/sha1.R
billdenney: I updated this.

Issue #140, opened 2019-12-04
Overhead in `digest()` due to `::`


wlandau:
As with `Sys.info()` in #137 and `match.arg()` in #139, `digest()` spends a noticeable amount of time on `::` (#138). How do you feel about using `serialze()` instead of `base::serialize()` if it removes this overhead?

eddelbuettel:
Dear Will.  Can we now stop piling new issues up here?  This is *fundamentally* all the same as #137, no?

eddelbuettel:
And what is `serialze()` ?

This package existed since R 1.8.0 brought us `base::serialize()`.  Relying on base R to serialize any odd R object has served us reasonably well.

wlandau:
> Dear Will. Can we now stop piling new issues up here? This is fundamentally all the same as #137, no?

Sure, sorry that was bothersome.

> And what is serialze()?

> This package existed since R 1.8.0 brought us base::serialize(). Relying on base R to serialize any odd R object has served us reasonably well.

Yes, I agree with using serialization from `base`. I just wondering about the (admittedly slight) cost of explicitly writing `base::`. `::` and `getExportedValue()` tend to appear in my profiling studies.

eddelbuettel:
There must have been a reason at some point why I explicitly prefixed it.  Maybe it was this:

```
2004-05-26  Dirk Eddelbuettel  <edd@debian.org>

        * Release 0.1.2
        - R/zzz.R: remove test for R < 1.8.0 and load of serialize package
          (as serialize has been removed from CRAN with serialize() in R)

```

wlandau:
Interesting, I did not know there was a `serialize` package originally. I see no danger of ambiguity now.

Perhaps I was excessive when I opened an issue about prefixing in `digest()`. However, I do think it could be an easy win for functions produced by `getVDigest()`. (By the way, thanks again for the tip in #139! It is already speeding up `drake`: https://github.com/ropensci/drake/issues/1092.)

```r
library(digest)
library(fs)
library(profile)
library(withr)

inner_digest <- getVDigest(algo = "murmur32")
digest_murmur32 <- function(object, serialize = TRUE, ...) {
  if (serialize) {
    inner_digest(list(object), serialize = TRUE, ...)
  } else {
    inner_digest(object, serialize = FALSE, ...)
  }
}

profile <- function(plan) {
  local_dir(dir_create(tempfile()))
  replicate(1e5, digest_murmur32(letters))
  rprof <- "prof.rprof"
[...]

PR #141, opened 2019-12-06
Cache nosharing state of `base::serialize()`

wlandau:
Re https://github.com/eddelbuettel/digest/issues/140#issuecomment-562138505, this PR pre-computes the value of `"nosharing" %in% names(formals(base::serialize))` for use in `digest()` and `non_streaming_digest()`. Benchmarks from https://github.com/eddelbuettel/digest/issues/140#issuecomment-562413844 show a 12% speedup in `digest()` and a 25% speedup in functions from `getVDigest()` in repeated calls on small data.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2019-12-05  Will Landau  <will.landau@gmail.com>
+
+	* R/init.R (.onLoad): Cache nosharing state in base::serialize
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Nice.  One most minor nit I can carry over is to maybe mention in init.R that the sharing only happens in the pqR variant by Radford.

PR #143, opened 2019-12-12
fix invalid code in documentation

aaronpeikert:
instead of "option" you likely wanted  "options"
and I think the version number needs quotes


eddelbuettel:
Nice catch, thank you!  You are correct on both counts.

Issue #136, opened 2019-11-22
sha1 tests fragility under r-devel on linux


eddelbuettel:
@billdenney @ThierryO 
I have a pending upload to CRAN but had to take out the tests deactivated in this commit:  https://github.com/eddelbuettel/digest/commit/ab60ab3c87e50a374b00376829fc8b488925bbf4

I was unable to confirm the failure here even though I have a current r-devel.  We will need to take another look.  It _may_ have to do with the changed treatment of matrices with respect to arrays, but that is just a guess...

billdenney:
I don't track r-devel closely as most of my work is relatively R version-independent.

With a quick look, I thin that you're correct for at least some of the tests.  the class for matrix objects will be `c("matrix", "array")` in 4.0.0 and future, so I think that at least some of the fixes may be:

Change:
https://github.com/eddelbuettel/digest/blob/ab60ab3c87e50a374b00376829fc8b488925bbf4/inst/tinytest/test_sha1.R#L99

to

`class = class(x.matrix.num)`

and change:

https://github.com/eddelbuettel/digest/blob/ab60ab3c87e50a374b00376829fc8b488925bbf4/inst/tinytest/test_sha1.R#L113

to

`class = class(x.matrix.letter)`

This line also has the same issue (though I think that the solution would be to precompute the hash for that matrix):
https://github.com/eddelbuettel/digest/blob/ab60ab3c87e50a374b00376829fc8b488925bbf4/inst/tinytest/test_sha1.R#L438

The other issues are less clear to me.

eddelbuettel:
Ok, I actually (just) had an r-devel version new enough but I did _not_ have the required environment variable set.  Once I set it the tests do fail for me.  I am not quite sure what the best course of action is but we can revisit this over the next few weeks -- no rush.

ThierryO:
My [git2rdata](https://github.com/ropensci/git2rdata) uses Travis to run checks on R release, oldrel and devel under Linux and release under OSx. Shall I send a PR to do this for digest too?

eddelbuettel:
Thanks for offering but I am very happy with the current Travis setup. It does what it does on purpose, and I maintain it over here: https://eddelbuettel.github.io/r-travis/  I have no issue running r-devel, the comment above was just a (very brief) follow-up explaining why my very first attempt at replicating failed.  It needed r-devel new enough (which I had) _and_ setting an env var (which I had not),

ThierryO:
[...]

PR #144, opened 2019-12-12
bugfix: ignore the srcref attribute

ThierryO:
For some reason, the `srcref` attribute changes when you store and read a function from a non-recommended package. Below is a repex using the latest CRAN version of digest.

```
f <- tempfile(fileext = ".rds")
x <- digest::sha1
saveRDS(x, f)
y <- readRDS(f)
options(sha1PackageVersion = NULL)
sha1(x) != sha1(y)
sha1(x, environment = FALSE) != sha1(y, environment = FALSE)
options(sha1PackageVersion = "0.6.22")
sha1(x) == sha1(y)
sha1(x, environment = FALSE) == sha1(y, environment = FALSE)
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2019-12-12  Thierry Onkelinx <thierry.onkelinx@inbo.be>
+
+	* DESCRIPTION (Version, Date): Roll minor version
+	* R/sha1.R: ignore the "srcref" attribute if set
+	* inst/tinytest/test_new_matrix_behaviour.R: update unit test
+	* inst/tinytest/test_sha1.R: update unit test
+	* man/sha1.Rd: update documentation
+
 2019-12-09  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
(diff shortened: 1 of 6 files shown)


ThierryO:
Still to do: update the ChangeLog and the documentation

ThierryO:
Maybe the attributes might also cause the issue in #136?

eddelbuettel:
(Can I bug you to maybe run a quick 'git rebase master' on your end? There is also a possible conflict, or may be git is smart enough to avoid -- your earliest commit in this overlaps with what I merged earlier)

image

eddelbuettel:
The other, bigger, question is:  will that not change behavior and tests?

ThierryO:
It is a breaking change for objects with a `srcref` attribute. The attribute is used in all functions, except those available in R-base. I'm not aware of other objects using that attribute.

I'll force push a new version.

eddelbuettel:
On balance this is likely "good enough" and we don't need to overcomplicate it -- so merging.

Thanks for the PR!

ThierryO:
> do we need to do "150%" and offer an option to maintain the old behavior.

That's why we have `options(sha1PackageVersion = "the version I prefer")` ;-)

Issue #146, opened 2020-02-06
spookyhash segfaults with instances of reference classes


LTLA:
Pretty much as the title says; a minimal working example might be:

```r
example(setRefClass) # use this to generate 'xx' instance of reference class.
digest::digest(xx, algo="spookyhash")
## 
##  *** caught segfault ***
## address 0x5, cause 'memory not mapped'
## 
## Traceback:
##  1: paste(.Call(spookydigest_impl, object, skip, 0, 0, serializeVersion),     collapse = "")
##  2: digest::digest(xx, algo = "spookyhash")
## 
## Possible actions:
## 1: abort (with core dump, if enabled)
## 2: normal R exit
## 3: exit R without saving workspace
## 4: exit R saving workspace
```

I stumbled across this when trying to hash a larger object that happened to contain an instance of a reference class (namely a `SummarizedExperiment` instance, if you're curious).

I can reproduce this on the following systems:

<details>
<summary>Linux, R 3.6.1</summary>

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: CentOS Linux 7 (Core)

Matrix products: default
BLAS:   /XXX/lib64/R/lib/libRblas.so
LAPACK: /XXX/lib64/R/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C
 [9] LC_ADDRESS=C               LC_TELEPHONE=C
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] digest_0.6.23

loaded via a namespace (and not attached):
[1] compiler_3.6.1
```
</details>

<details>
<summary>MacOSX, R-devel</summary>

```
R Under development (unstable) (2019-11-04 r77366)
Platform: x86_64-apple-darwin17.7.0 (64-bit)
Running under: macOS High Sierra 10.13.6

Matrix products: default
BLAS:   /XXX/lib/libRblas.dylib
LAPACK: /XXX/lib/libRlapack.dylib

locale:
[...]

PR #148, opened 2020-02-17
add back fun=NULL

kendonB:
Solves #146. Was a dumb mistake in the end where there is a `fun` argument to `spookydigest_impl` which doesn't get put in.

```r
> temp_library_location <- tempfile(pattern="lib")
> dir.create(temp_library_location)
> .libPaths(temp_library_location)
> install.packages("https://github.com/kendonB/digest/archive/ecb34bc.tar.gz", type = "source", repos = NULL, quiet = TRUE)
trying URL 'https://github.com/kendonB/digest/archive/ecb34bc.tar.gz'
downloaded 126 KB

> example(setRefClass, echo = FALSE) # use this to generate 'xx' instance of reference class.
Call:
$undo()


Undoes the last edit() operation
        and update the edits field accordingly.
        

Reference class 'mEdit' [package ".GlobalEnv"] with 2 fields
 $ data : num [1:4, 1:3] 1 2 3 4 5 6 7 8 9 10 ...
 $ edits: list()
 and 17 methods, of which 3 are  possibly relevant:
   edit, show#envRefClass, undo
> # Works
> digest::digest(xx, algo = "spookyhash")
[1] "d4fbf9c7699d4303503d108c5722d7af"
```

### DIFF
--- a/R/digest.R
+++ b/R/digest.R
@@ -108,7 +108,7 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                      as.integer(seed))
     } else if (algo == "spookyhash"){
         # 0s are the seeds. They are included to enable testing against fastdigest.
-        val <- paste(.Call(spookydigest_impl, object, skip, 0, 0, serializeVersion), collapse="")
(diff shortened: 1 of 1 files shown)


eddelbuettel:
> Was a dumb mistake in the end

Taadaa!!  Been there, done that....

Issue #49, opened 2016-11-11
could you replace `file.access` with `R.utils::fileAccess` ?


stla:
Hello, 

This is a message for the author of the `digest` package.

It is known that `file.access` has some issues with network drives, as we can see [in this issue](https://github.com/eddelbuettel/digest/issues/13) and [in this post on stackoverflow](http://stackoverflow.com/q/40530968/1100107).

I am the author of this post on stackoverflow. I get:
```
> file.access("U:/Data", 4)
U:/Data 
     -1 
```
I have tried the `fileAccess` function from the `R.utils` package:
```
> R.utils::fileAccess("U:/Data", 4)
[1] 0
Warning message:
In fileAccess.default("U:/Data", 4) :
  file.access(..., mode=4) and file(..., open="rb")+readBin() gives different results (-1 != 0). Will use the file()+readBin() results: U:/Data
```
This time, I get `0`. 
Thus, my problem (and perhaps other problems) would be solved if you replaced `file.access` with `R.utils::fileAccess`.

eddelbuettel:
No, I can't / won't because it would add another dependency needed only by
- a subset of OSs
- a subset of users with wrongly configured network drives
and proliferation of dependencies is generally not a desirable thing.  Rest assured that millions of users get digest to work faithfully for them, including on Windows. And network drives.

Now, if you really must change the code and cannot change your directory, then I would consider a _really careful_ pull request against digest which branches the execution and possibly doubles the code from R.utils.  @HenrikBengtsson is already a contributor and co-author and won't mind (I surmise...)

stla:
Ok, thank you for your attention and your reply. 
I find very strange that `file.access` returns `-1` whereas the folder *is* readable. Could one consider that this is a bug of the function `file.access` ? In this case, this would be a bug to be reported to the R Core Team (https://www.r-project.org/bugs.html). But maybe I misunderstand something and the returned value `-1` is the right value to be returned....

stla:
> users with wrongly configured network drives
Indeed, this is possibly a wrong configuration... I'll try to talk about that to the IT team of my company but I'm afraid it will be hard to explain ^^

stla:
I've found a workaround.
```
mydigest <- function (object, algo = c("md5", "sha1", "crc32", "sha256", 
    "sha512", "xxhash32", "xxhash64", "murmur32"), serialize = TRUE, 
    file = FALSE, length = Inf, skip = "auto", ascii = FALSE, 
[...]

Issue #149, opened 2020-02-29
hope to address file.access issue, without adding dependencies


ashlinrichardson:
Dear Eddel Buettel,
Sir:

Thank you for your excellent software! @boshek pointed out that at line 170 of
https://github.com/eddelbuettel/digest/blob/master/R/digest.R

there's a potential minor improvement, since file.access() isn't reliable for distributed systems

At https://github.com/eddelbuettel/digest/issues/49
you noted that adding more dependencies would not be a solution at all-- and I would certainly agree with your view.

Would you be open to a PR that would implement the minor improvement, without adding any dependencies?

Thanks for considering this

Many thanks
cheers,

Ashlin Richardson
Senior Data Scientist
Government of British Columbia
Victoria, BC Canada

eddelbuettel:
Hi @Ashlin, 

What exactly are you suggesting? If it is this line
https://github.com/eddelbuettel/digest/blob/b0f04b4f67865f79e01253b351f009db8495e582/R/digest.R#L170
then I have to refer you to R itself as we discussed in #49 -- and eg confirmed by @HenrikBengtsson .  This is not a shortcoming in the `digest` package, so it should preferably be addressed where it belongs, in R itself.  Or in your application layer: if the shared drive does not support POSIX semantics, copy the file to your local machine and its filesystem and then call `digest`.

What would be the "fix" for this you are contemplating?  If it is simple and small  we can consider it if it doesn't have side effects -- though I still think "fixing symptoms" is bad/wrong.  

Dirk

ashlinrichardson:
@eddelbuettel 

Hi Dirk, 
Yes, agreed--
--not hoping to mask symptoms of a problem that's in R,

rather I was hoping to align with the spirit of your extensive work on the Rcpp and inline packages, by perhaps getting a little closer to the horse's mouth:

by using an R call that's slightly more direct or reliable:
        file()

Wanting to develop things like Rcpp shows great care, attention to detail and perseverance,  so I thought a change to a lower-level function that's more stable across platforms, might be of interest

Will attach a PR in moment to illustrate-- even if you like the idea -- you're likely to be able to express it better or more concisely than I am
cheers,

Ash

eddelbuettel:
No don't attach PR.  Bring is *here* because *here* is where we have the discussion.

And maybe also try to address the _qualitative_ issue from the prior (closed !!) issue on the _very same topic_.  So far you have not addded anything new to the debate.

[...]

PR #150, opened 2020-02-29
Cleanly sidestep file.access issue, to add reliability without adding deps

ashlinrichardson:
Hi Dirk,
Attempt at cleanly sidestepping (not fixing) file.access issue,  https://github.com/eddelbuettel/digest/issues/49
without creating any dependencies,
and without masking a problem that's within R.

Hoping to align with the spirit of your extensive work on the Rcpp and inline packages, by perhaps getting a little closer to the horse's mouth:

by using an R call that's slightly more direct or reliable:
file()

Wanting to develop things like Rcpp shows great care, attention to detail and perseverance, so I thought a change to a lower-level function that's more stable across platforms, might be of interest

This PR attached to illustrate. Even if you like the idea -- you're likely to be able to express it better or more concisely than I am! Thanks for considering the idea
cheers,

### DIFF
--- a/R/digest.R
+++ b/R/digest.R
@@ -167,7 +167,9 @@ check_file <- function(object, errormode){
         return(.errorhandler("The specified pathname is not a file: ",
                              object, mode=errormode))
     }
-    if (file.access(object, 4)) {
+    read_test <- NULL
+    try(read_test <- file(object, "rb"))
+    if(typeof(read_test) != "integer"){
         return(.errorhandler("The specified file is not readable: ",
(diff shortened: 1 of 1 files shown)


eddelbuettel:
With _all_ due respect this is not helpful.  You opened the issue hours ago, I asked you a question there and we had _not yet concluded_ a PR was helpful.  

I don't consider this helpful yet.  I am also away from home at an R conference you that expect within the hour (or sooner) replies from an unpaid volunteer is a little heavy.

eddelbuettel:
Closing this, see #149 for comments.

PR #151, opened 2020-03-05
Update sha1.Rd

harrismcgehee:
Fixed typo


eddelbuettel:
Thanks for catching that!

PR #164, opened 2020-10-07
Speed enhancements

### LINKED ISSUE #163: Possible performance improvements
I'm doing some work on the memoise package, and have found that the performance of `digest()` could be improved with some simple changes.

If the input to `digest()` is large, then the bottleneck is the hashing algorithm (which is something I don't want to mess with). However, when the input is small, then the bottleneck is miscellaneous in the R code, some of which can be easily improved.

When a user calls a memoized function with small inputs, the overhead from this R code can be significant. I've put a profile of a run here:
https://rpubs.com/wch/671614

To reproduce a similar profile, you need to be running R 3.6 (there's a bug in the profiler in R 4.0 that makes it not record the top of the call stack, rendering it almost useless).

```R
# Install digest with source refs
install.packages("digest", INSTALL_opts = "--with-keep.source", type = "source", 
                 repos = "https://cloud.r-project.org/")

library(digest)
profvis::profvis({
  for (i in 1:20000) {
    digest(123)
  }
})
```

The total time is 590ms. Of that:

* 180ms is spent in `match.arg()`
* 80ms is spent in `::` (for `base::serialize`)

So 44% of the time spent in that code, and those can easily be optimized. I'm happy to provide a PR if that would help.

wch:
Closes #163.

The changes:

* Replaced `base::serialize` with `serialize`.
* Pass explicit choices to `match.arg()`.
* Only check if an algorithm is streaming one time.
* Faster check for `skip=="auto"`. The previous code for that used `any(!is.na(pmatch(skip,"auto")))`. That was added a long time ago, in 543df6705591e1eea585e829a2f0ba26fedff1ea, and it seems unnecessarily complicated. The new version checks for an exact match; the previous version allowed partial matches, like if `skip` was `"au"`, but that doesn't seem like desirable behavior to me.

There's still a little room for improvement by having a replacement for `match.arg`, but I decided against it after I tried it out and found that the benefits were marginal.

Simple benchmark results:

```R
library(digest)
microbenchmark::microbenchmark(
  digest(123, algo = "sha512"),
  times = 10000
)

## Before
#> Unit: microseconds
#>                          expr    min      lq    mean median     uq     max neval
#>  digest(123, algo = "sha512") 34.839 43.5675 46.8873 45.373 47.695 136.589  1000

## After
#> Unit: microseconds
#>                          expr    min      lq     mean median     uq     max neval
#>  digest(123, algo = "sha512") 19.817 20.5795 22.08357 21.029 21.849 108.512  1000
```

And some profiles:

Before (590ms): https://rpubs.com/wch/671614
After (280ms): https://rpubs.com/wch/671745

Interestingly, in the After condition, serialization took 30% of the total time and the actual hashing took about 18% of the total time.

### DIFF
--- a/R/digest.R
+++ b/R/digest.R
@@ -36,8 +36,12 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                    errormode=c("stop","warn","silent"),
                    serializeVersion=.getSerializeVersion()) {
 
-    algo <- match.arg(algo)
-    errormode <- match.arg(errormode)
+    # Explicitly specify choices; this is much faster than having match.arg()
+    # infer them from the function's formals.
+    algo <- match.arg(algo, c("md5", "sha1", "crc32", "sha256", "sha512",
+                              "xxhash32", "xxhash64", "murmur32",
+                              "spookyhash", "blake3"))
+    errormode <- match.arg(errormode, c("stop", "warn", "silent"))
 
     if (is.infinite(length)) {
         length <- -1               # internally we use -1 for infinite len
--- a/R/digest.R
+++ b/R/digest.R
@@ -58,12 +58,12 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
 
     if (serialize && !file) {
         if(algo %in% non_streaming_algos){
-            ## support the 'nosharing' option in pqR's base::serialize()
+            ## support the 'nosharing' option in pqR's serialize()
             object <- if (.hasNoSharing())
-                base::serialize (object, connection=NULL, ascii=ascii,
+                serialize (object, connection=NULL, ascii=ascii,
                                  nosharing=TRUE, version=serializeVersion)
             else
-                base::serialize (object, connection=NULL, ascii=ascii,
+                serialize (object, connection=NULL, ascii=ascii,
                                  version=serializeVersion)
         }
         ## we support raw vectors, so no mangling of 'object' is necessary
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Thanks for that -- looks good at fist glance (apart of course from the continued horror that are `snake_case` and lack of spaces after `if` and before `{` but you gotta do ... ;-)

I'll give it another look tomorrow in a quiet moment.

wch:
Oh ha, the lack of spaces was because I was just trying to copy the existing style of that code... I've fixed it.

eddelbuettel:
I was mostly pulling your leg here.  (But the one thing I do dislike quite actively after 20+ years of ESS and R Core styling on R code is what your mothership now seems to impose.  Not a fan.  The spaces ... are fine.  I am used to lots of things from numerous collaborators.  

I'll try to get to it today.

eddelbuettel:
I'll merge now. I am a bit surprised we allow ourselves the microoptimisation of copying possible value for `match.arg()` from `formals()` as it opens the door to a possible 'gap' the function interface ever change.  Given the maturity of the package that is not that certain but it is still a little ... awkward.

But I respect your empirics -- that is nice work so thank you!

wch:
FWIW, the unit tests cover the possible "gap" between the formals and the `choices` passed to `match.arg`. If someone adds a new algorithm but doesn't add it to the `choices`, then the test will fail. (Although, presumably if someone is adding a new algorithm, they'll actually test it out in development and find then that it's not getting past the `match.arg()`.

eddelbuettel:
Penny just dropped as I just asked myself "but how".  Yes indeed because you outsource the read of the formals to the test. 

That is a nice one!

PR #168, opened 2021-03-22
digest.Rd: minor documentation updates

florisvdh:
Some suggestions:

- xxhash and others are non-cryptographic hash functions, so better not refer to cryptographic functions only (in the description)
- I suggest to advertise more explicitly that `digest` can also be used to generate file hashes.

### DIFF
--- a/man/digest.Rd
+++ b/man/digest.Rd
@@ -2,7 +2,7 @@
 \alias{digest}
 \title{Create hash function digests for arbitrary R objects}
 \description{
-  The \code{digest} function applies a cryptographical hash function to
+  The \code{digest} function applies a hash function to
   arbitrary R objects. By default, the objects are internally
   serialized, and either one of the currently implemented MD5 and SHA-1
   hash functions algorithms can be used to compute a compact digest of
--- a/man/digest.Rd
+++ b/man/digest.Rd
@@ -1,9 +1,9 @@
 \name{digest}
 \alias{digest}
-\title{Create hash function digests for arbitrary R objects}
+\title{Create hash function digests for arbitrary R objects or files}
 \description{
   The \code{digest} function applies a hash function to
-  arbitrary R objects. By default, the objects are internally
+  arbitrary R objects or files. By default, the objects are internally
   serialized, and either one of the currently implemented MD5 and SHA-1
   hash functions algorithms can be used to compute a compact digest of
   the serialized object.


eddelbuettel:
> xxhash and others are non-cryptographic hash functions, so better not refer to cryptographic functions only (in the description)

I am not sure I understand what you are trying to say. Can you reword?  `digest` has said since the very beginning that it is NOT for cryptographic purposes but for hash digests.  In the DESCRIPTION file we say

     Please note that this package is not meant to be deployed for cryptographic purposes for which more
     comprehensive (and widely tested) libraries such as 'OpenSSL' should be used.

I was hoping this to be clear enough.  Are you suggesting it is not?   What should we write instead?

man/digest.Rd
```diff
@@ -1,9 +1,9 @@
 \name{digest}
 \alias{digest}
-\title{Create hash function digests for arbitrary R objects}
+\title{Create hash function digests for arbitrary R objects or files}
```
eddelbuettel, line 3: I am on the fence here as titles should be short and succinct.  I will probably merge the PR but may remove these two words later -- will ponder.

man/digest.Rd
```diff
@@ -1,9 +1,9 @@
...
 \description{
-  The \code{digest} function applies a cryptographical hash function to
```
eddelbuettel, line 5: Good catch to remove that.  Might be worth mentioning that the cryptography context is where the hash functions come from, but as it is now it is misleading.

man/digest.Rd
```diff
@@ -1,9 +1,9 @@
...
-  arbitrary R objects. By default, the objects are internally
+  The \code{digest} function applies a hash function to
+  arbitrary R objects or files. By default, the objects are internally
```
eddelbuettel, line 6: The 'or files'  is perfect here -- thanks.

eddelbuettel (approved):

florisvdh:
@eddelbuettel OK! I'm no expert in this at all, but I think this provides a good overview of 'cryptographic' vs. 'non-cryptographic' hash functions: https://en.wikipedia.org/wiki/List_of_hash_functions

With regard to the title, that may be personal preference indeed. I was more interested in file hashing - in the long term, R object hashes seem more subject to software changes (R / package versions). Initially I overlooked `digest` for file hashing because I thought (from its description etc) it was for R objects only. That's why I also suggested ce36f4f. Of course, decision is up to you.

eddelbuettel:
> in the long term, R object hashes seem more subject to software changes (R / package versions).

Could not disagree any more strongly here.  The package is turning eighteen years this year.  There was one change in R format.

man/digest.Rd
eddelbuettel: I will reword this in a way to say that one of many available hash functions is applied to a given object.  Not sure yet why, and as you point out these functions are both cryptographic and non-cryptographic.

florisvdh:
> R / package versions

I could be have been clearer. I was not referring to `digest`, but just to R packages which generate R objects, which one may want to hash with `digest`. If an R function (in some package) changes the structure of its output in a new version of that package, the object hash will change. That may or may not be relevant, depending on the usecase, i.e. what exactly needs to be verified.

My usecase is to verify input data integrity (of reference data sets), therefore I find it more appropriate (long-term stability) to check the file directly, not its result after reading in R because the latter may change with time.

florisvdh:
> Detecting that change is a feature

I know that, and this capability is really great - not shooting anyone... I just tried to describe my usecase, where file versions are to be checked rather than R objects. That's all, really...

man/digest.Rd
florisvdh: OK

eddelbuettel:
Thanks again -- I added a ChangeLog entry and reworded that opening paragraph just a tiny bit more (and reindented).

florisvdh:
Thank you!

PR #175, opened 2021-12-15
Add Loongarch info

fangyaling:
Add loongarch info

### DIFF
--- a/src/pmurhash.c
+++ b/src/pmurhash.c
@@ -123,7 +123,7 @@ on big endian machines, or a byte-by-byte read if the endianess is unknown.
 
 /* gcc (usually) defines xEL/EB macros for ARM and MIPS endianess */
 #if !defined(__BYTE_ORDER)
-  #if defined(__ARMEL__) || defined(__MIPSEL__)
+  #if defined(__ARMEL__) || defined(__MIPSEL__) || defined(__loongarch__)
     #define __BYTE_ORDER __LITTLE_ENDIAN
   #endif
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Sure, can do -- can you remind me (for documentation purposes) which CPU types identify that way?

fangyaling:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has Reduced Instruction Set Computer (RISC) style. And more infomation you can get from the link:https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html

eddelbuettel:
Very good, thanks/. Almost too much detail.  

Could you maybe provide simpler information on
- what a common cpu name is
- who manufactures these 
- what brand names they may have
- what OS is typically running

to help me assess relevancy?

eddelbuettel:
This helps somewhat:  https://en.wikipedia.org/wiki/Loongson

So is this a mostly / exclusively within-China Linux-style cpu?

fangyaling:
Thanks for review. But now I will close this change for duplicate definition after checking the gcc define.

eddelbuettel:
Oh, so the `gcc` define works for you?  Otherwise we can of course enable this.  I am just not that familiar with non-standard cpus (and for me that means non-Intel or AMD, I just don't keep track of the many ARM or MIPS variants ...).   

Let me know if this needs anything, if so we should support it.

fangyaling:
OK. I got it.Thanks!

Issue #173, opened 2021-10-29
Consider meow


eddelbuettel:
Cf https://mollyrocket.com/meowhash

> ## How fast is it?
> It’s the fastest hash function we know of, and we have benchmarked all the ones we could find. On modern Intel x64 CPUs, it hashes 16 bytes per cycle single-threaded. This means in cache it can hash at a rate of 64 gigabytes per second on a 4.2gHz machine. Out of cache, it hashes at whatever speed your main memory bus can provide to a single core, since that is usually the limiting factor on modern x64 CPUs.
>
> It has also now been tuned to be the fastest hash on small inputs, too. Despite the fact that it is a full 128-bit hash, it still outperforms “fast” 64-bit hashes across all input sizes. So whether you have data buffers of a few bytes or a few gigabytes, Meow is the fastest hash available.

eddelbuettel:
Or maybe not -- see (extended) discussion in #180.

PR #180, opened 2022-10-06
[Experiment] Investigate meow, digest benchmarks

nx10:
I had a look at meow (see #173) as a possible hacktoberfest contribution, but it did not go as planned. This PR is a minimal working example of meow (no file support, no docs, no configure script for different processors / compilers) and not intended for merge, however I want to share some observations I made with this:

First of all meow does require x64 as well as hardware AES, so I assume it's not a good fit for `digest` over all.

However to see how fast it is I created a small benchmark:

```R
ben <- function(object, num_reps, serialize = TRUE) {

    df <- rbenchmark::benchmark(
        "md5" = {
            digest::digest(object, "md5", serialize = serialize)
        },
        "sha1" = {
            digest::digest(object, "sha1", serialize = serialize)
        },
        "crc32" = {
            digest::digest(object, "crc32", serialize = serialize)
        },
        "sha512" = {
            digest::digest(object, "sha512", serialize = serialize)
        },
        "sha1" = {
            digest::digest(object, "sha1", serialize = serialize)
        },
        "xxhash32" = {
            digest::digest(object, "xxhash32", serialize = serialize)
        },
        "xxhash64" = {
            digest::digest(object, "xxhash64", serialize = serialize)
        },
        "murmur32" = {
            digest::digest(object, "murmur32", serialize = serialize)
        },
        "spookyhash" = {
            digest::digest(object, "spookyhash", serialize = TRUE)  # serialize=F not supported
        },
        "blake3" = {
            digest::digest(object, "blake3", serialize = serialize)
        },
        "meow" = {
            digest::digest(object, "meow", serialize = serialize)
        },
        replications = num_reps, order = "elapsed")

    df$hash_width <- lapply(df$test, \(x) nchar(digest::digest(0, x)) * 4)
    df$bandwidth <- sprintf("%.2f GB/s", object.size(object) * num_reps / 1024^3 / df$elapsed)
    df
}
```

I ran the following tests on my Windows 10, R 4.2, with a 12th Gen Intel(R) Core(TM) i7-12700K, 3610 Mhz, 12 Core(s), 20 Logical Processor(s).

<details>
  <summary>A small size (100kB) test with many reps</summary>

```R
ben(as.raw(runif(100 * 1024)), num_reps = 10000, serialize = T) # |> knitr::kable(row.names = F)
```

|test       | replications| elapsed| relative| user.self| sys.self| user.child| sys.child|hash_width |bandwidth |
|:----------|------------:|-------:|--------:|---------:|--------:|----------:|---------:|:----------|:---------|
|spookyhash |        10000|    0.18|    1.000|      0.18|     0.00|         NA|        NA|128        |5.30 GB/s |
|meow       |        10000|    0.42|    2.333|      0.27|     0.16|         NA|        NA|128        |2.27 GB/s |
|xxhash64   |        10000|    0.48|    2.667|      0.33|     0.16|         NA|        NA|64         |1.99 GB/s |
|xxhash32   |        10000|    0.51|    2.833|      0.31|     0.20|         NA|        NA|32         |1.87 GB/s |
|murmur32   |        10000|    0.64|    3.556|      0.44|     0.20|         NA|        NA|32         |1.49 GB/s |
|crc32      |        10000|    1.00|    5.556|      0.80|     0.20|         NA|        NA|32         |0.95 GB/s |
|blake3     |        10000|    1.33|    7.389|      1.24|     0.09|         NA|        NA|256        |0.72 GB/s |
|sha1       |        10000|    1.34|    7.444|      1.09|     0.25|         NA|        NA|160        |0.71 GB/s |
|sha1       |        10000|    1.40|    7.778|      1.27|     0.13|         NA|        NA|160        |0.68 GB/s |
|md5        |        10000|    1.60|    8.889|      1.36|     0.22|         NA|        NA|128        |0.60 GB/s |
|sha512     |        10000|    2.28|   12.667|      2.00|     0.27|         NA|        NA|512        |0.42 GB/s |

</details>

<details>
  <summary>A larger (1GB) test with 1 rep</summary>

```R
ben(as.raw(runif(1024^3)), num_reps = 1, serialize = T) # |> knitr::kable(row.names = F)
```

|test       | replications| elapsed| relative| user.self| sys.self| user.child| sys.child|hash_width |bandwidth  |
|:----------|------------:|-------:|--------:|---------:|--------:|----------:|---------:|:----------|:----------|
|spookyhash |            1|    0.08|    1.000|      0.08|     0.00|         NA|        NA|128        |12.50 GB/s |
|meow       |            1|    1.36|   17.000|      1.00|     0.36|         NA|        NA|128        |0.74 GB/s  |
|xxhash64   |            1|    1.39|   17.375|      0.84|     0.55|         NA|        NA|64         |0.72 GB/s  |
|xxhash32   |            1|    1.41|   17.625|      0.86|     0.54|         NA|        NA|32         |0.71 GB/s  |
|murmur32   |            1|    1.56|   19.500|      1.09|     0.47|         NA|        NA|32         |0.64 GB/s  |
|crc32      |            1|    1.86|   23.250|      1.41|     0.46|         NA|        NA|32         |0.54 GB/s  |
|sha1       |            1|    2.18|   27.250|      1.64|     0.55|         NA|        NA|160        |0.46 GB/s  |
|sha1       |            1|    2.19|   27.375|      1.91|     0.28|         NA|        NA|160        |0.46 GB/s  |
|blake3     |            1|    2.27|   28.375|      1.83|     0.44|         NA|        NA|256        |0.44 GB/s  |
|md5        |            1|    2.42|   30.250|      2.03|     0.39|         NA|        NA|128        |0.41 GB/s  |
|sha512     |            1|    3.15|   39.375|      2.75|     0.39|         NA|        NA|512        |0.32 GB/s  |
 
</details>

meow was a bit faster than most other algorithms except spookyhash, which completely blows everything else out of the water, especially with big objects, where it was 17x-39x faster.

To check if there was something going on with serialization, I wanted to run the test again with serialization disabled, which I then found out was not supported in combination with spookyhash. 

Nonetheless here are the results with `serialize=F` for all algorithms but spookyhash:

<details>
  <summary>100kB - no serialze</summary>

```R
ben(as.raw(runif(100 * 1024)), num_reps = 10000, serialize = F) # |> knitr::kable(row.names = F)
```

|test       | replications| elapsed| relative| user.self| sys.self| user.child| sys.child|hash_width |bandwidth |
|:----------|------------:|-------:|--------:|---------:|--------:|----------:|---------:|:----------|:---------|
|meow       |        10000|    0.10|      1.0|      0.09|     0.00|         NA|        NA|128        |9.54 GB/s |
|xxhash64   |        10000|    0.13|      1.3|      0.13|     0.00|         NA|        NA|64         |7.34 GB/s |
|xxhash32   |        10000|    0.17|      1.7|      0.17|     0.00|         NA|        NA|32         |5.61 GB/s |
|spookyhash |        10000|    0.18|      1.8|      0.17|     0.00|         NA|        NA|128        |5.30 GB/s |
|murmur32   |        10000|    0.31|      3.1|      0.28|     0.00|         NA|        NA|32         |3.08 GB/s |
|crc32      |        10000|    0.62|      6.2|      0.63|     0.00|         NA|        NA|32         |1.54 GB/s |
|sha1       |        10000|    0.98|      9.8|      0.99|     0.00|         NA|        NA|160        |0.97 GB/s |
|blake3     |        10000|    0.99|      9.9|      0.98|     0.00|         NA|        NA|256        |0.96 GB/s |
|sha1       |        10000|    1.03|     10.3|      1.00|     0.02|         NA|        NA|160        |0.93 GB/s |
|md5        |        10000|    1.13|     11.3|      1.13|     0.00|         NA|        NA|128        |0.84 GB/s |
|sha512     |        10000|    1.73|     17.3|      1.74|     0.00|         NA|        NA|512        |0.55 GB/s |

(spookyhash serialize=T)
</details>

<details>
  <summary>1GB - no serialze</summary>
  
```R
ben(as.raw(runif(1024^3)), num_reps = 1, serialize = F) # |> knitr::kable(row.names = F)
```
  
|test       | replications| elapsed| relative| user.self| sys.self| user.child| sys.child|hash_width |bandwidth  |
|:----------|------------:|-------:|--------:|---------:|--------:|----------:|---------:|:----------|:----------|
|meow       |            1|    0.03|    1.000|      0.03|        0|         NA|        NA|128        |33.33 GB/s |
|xxhash64   |            1|    0.08|    2.667|      0.08|        0|         NA|        NA|64         |12.50 GB/s |
|spookyhash |            1|    0.08|    2.667|      0.07|        0|         NA|        NA|128        |12.50 GB/s |
|xxhash32   |            1|    0.11|    3.667|      0.11|        0|         NA|        NA|32         |9.09 GB/s  |
|murmur32   |            1|    0.23|    7.667|      0.23|        0|         NA|        NA|32         |4.35 GB/s  |
|crc32      |            1|    0.56|   18.667|      0.56|        0|         NA|        NA|32         |1.79 GB/s  |
|sha1       |            1|    0.93|   31.000|      0.94|        0|         NA|        NA|160        |1.08 GB/s  |
|blake3     |            1|    0.95|   31.667|      0.95|        0|         NA|        NA|256        |1.05 GB/s  |
|sha1       |            1|    0.96|   32.000|      0.95|        0|         NA|        NA|160        |1.04 GB/s  |
|md5        |            1|    1.10|   36.667|      1.10|        0|         NA|        NA|128        |0.91 GB/s  |
|sha512     |            1|    1.78|   59.333|      1.78|        0|         NA|        NA|512        |0.56 GB/s  |
  
(spookyhash serialize=T)
</details>

Interestingly spookyhash with serialization seems to hold itself up against xxhash and meow even when disabling serialization for the other two (although meow is 3x faster with the large data).

I think there is something strange going on with spookyhash. I noticed in the code that it is treated differently than the other hash functions. Is it possible that the spookyhash implementation somehow avoids a copy? Can we avoid this copy in the other algorithms to make digest substantially faster? Or is the spookyhash implementation broken?

Of course all of this could also just be the result of an error I made.

Some closing points:
- Possibly unrelated, but spookyhash does also not seem to respect `raw=TRUE`
- For reference an independent hash performance benchmark: https://github.com/Cyan4973/xxHash/wiki/Performance-comparison 

Cheers,
Florian

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,2 +1,3 @@
 
 CXX_STD = CXX11
+PKG_CPPFLAGS = -mavx -maes
--- a/R/digest.R
+++ b/R/digest.R
@@ -40,7 +40,7 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
     # infer them from the function's formals.
     algo <- match.arg(algo, c("md5", "sha1", "crc32", "sha256", "sha512",
                               "xxhash32", "xxhash64", "murmur32",
-                              "spookyhash", "blake3"))
+                              "spookyhash", "blake3", "meow"))
     errormode <- match.arg(errormode, c("stop", "warn", "silent"))
 
     if (is.infinite(length)) {
@@ -149,7 +149,8 @@ algo_int <- function(algo)
         xxhash64 = 7,
         murmur32 = 8,
         spookyhash = 9,
-        blake3 = 10
+        blake3 = 10,
+        meow = 11
     )
 
 ## HB 14 Mar 2007:
--- a/R/digest.R
+++ b/R/digest.R
@@ -25,7 +25,7 @@
 
 digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                                   "xxhash32", "xxhash64", "murmur32",
-                                  "spookyhash", "blake3"),
+                                  "spookyhash", "blake3", "meow"),
                    serialize=TRUE,
                    file=FALSE,
                    length=Inf,
--- a/src/Makevars.win
+++ b/src/Makevars.win
@@ -1,0 +1,1 @@
+PKG_CFLAGS = -mavx -maes
--- a/src/digest.c
+++ b/src/digest.c
@@ -37,6 +37,7 @@
 #include "xxhash.h"
 #include "pmurhash.h"
 #include "blake3.h"
+#include "meow.h"
 
 #ifdef _WIN32
 #include <Windows.h>
@@ -228,6 +229,16 @@ SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Se
             for (size_t i = 0; i < BLAKE3_OUT_LEN; i++) {
                 sprintf(output + i * 2, "%02x", val[i]);
             }
+        }
+        break;
+    }
+    case 11: {     /* meow */
+        output_length = sizeof(meow_u128);
+        meow_u128 val = MeowHash(MeowDefaultSeed, nChar, txt);
+        if (leaveRaw) {
+            memcpy(&output, &val, sizeof(meow_u128));
+        } else {
+            sprintf(output, "%08x%08x%08x%08x", MeowU32From(val, 3), MeowU32From(val, 2), MeowU32From(val, 1), MeowU32From(val, 0));
         }
         break;
     }
--- a/src/meow.h
+++ b/src/meow.h
@@ -1,0 +1,744 @@
+/* ========================================================================
+
+   Meow - A Fast Non-cryptographic Hash
+   (C) Copyright 2018-2019 by Molly Rocket, Inc. (https://mollyrocket.com)
+   
+   See https://mollyrocket.com/meowhash for details.
+   
+   ========================================================================
+   
+   zlib License
+   
+   (C) Copyright 2018-2019 Molly Rocket, Inc.
+   
+   This software is provided 'as-is', without any express or implied
+   warranty.  In no event will the authors be held liable for any damages
+   arising from the use of this software.
+   
+   Permission is granted to anyone to use this software for any purpose,
+   including commercial applications, and to alter it and redistribute it
+   freely, subject to the following restrictions:
+   
+   1. The origin of this software must not be misrepresented; you must not
+      claim that you wrote the original software. If you use this software
+      in a product, an acknowledgment in the product documentation would be
+      appreciated but is not required.
+   2. Altered source versions must be plainly marked as such, and must not be
+      misrepresented as being the original software.
+   3. This notice may not be removed or altered from any source distribution.
+   
+   ========================================================================
+   
+   FAQ
+   
+   Q: What is it?
+   
+   A: Meow is a 128-bit Level 3 hash taking 128 bytes of seed.  It operates
+      at very high speeds on x64 processors, and potentially other processors
+      that provide accelerated AES instructions.
+      
+   Q: What is it GOOD for?
+   
+   A: Quickly hashing any amount of data for comparison purposes such as
+      block deduplication or change detection.  It is fast on all buffer
+      sizes, and can generally be used anywhere you need fast Level 3
+      hashing without worrying about how big or small the inputs tend to be.
+      
+      However, substantial speed improvements could be made over Meow
+      if you either a) know you are always hashing an exact, small number of bytes,
+      or b) can always supply a small number of bytes in a buffer padded to some
+      fixed multiple of 16.
+      
+   Q: What is it BAD for?
+   
+   A: Anything requiring Level 4 or Level 5 security guarantees (see
+      http://nohatcoder.dk/2019-05-19-1.html#level3).  Also, note that
+      Meow is a new hash and has not had the extensive community
+      cryptanalysis necessary to ensure that it is not breakable down to
+      a lower level of hash, so you must do your due diligence in
+      deciding when and where to use Meow instead of a slower but
+      more extensively studied existing hash.  We have tried to design
+      it to provide Level 3 security, but the possibility of the hash
+      being broken in the future always exists.
+      
+   Q: Why is it called the "Meow hash"?
+   
+   A: It is named after a character in Meow the Infinite
+      (https://meowtheinfinite.com)
+      
+   Q: Who wrote it?
+   
+   A: The final Meow Hash was created as a collaboration between
+      JACOB CHRISTIAN MUNCH-ANDERSEN (https://twitter.com/nohatcoder) and
+      CASEY MURATORI (https://caseymuratori.com).  Casey wrote the original
+      implementation for use in processing large-footprint assets for the
+      game 1935 (https://molly1935.com).  Jacob was the first to analyze
+      that implementation and determine the adversarial bit strength, which
+      was weaker than they would have liked.
+      
+      Following that, the two collaborated to figure out how the hash
+      could be strengthened without reducing Meow's 16 bytes/cycle
+      maximum theoretical throughput.  Jacob created the hash candidates
+      and Casey did the performance validation.  After a long and
+      exhaustive effort, Jacob found the unaligned aes/add/xor formulation
+      that forms the current Meow hash core.
+      
+      A number of valuable additions to Meow Hash were also contributed
+      by other great folks along the way:
+      
+      JEFF ROBERTS (https://radgametools.com) provided a super slick
+      way to handle the residual end-of-buffer bytes that dramatically
+      improved Meow's small hash performance.
+      
+      MARTINS MOZEIKO (https://matrins.ninja) ported Meow to ARM and
+      ANSI-C, and added the proper preprocessor dressing for clean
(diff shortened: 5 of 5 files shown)


nx10:
Thanks!

> Maybe as a vignette?

Do you mean adding the benchmark as a vignette in a separate PR without meow? I would be happy to!

> any idea why CI failed?

Sorry about that, I missed adding meow to the default arguments. The tests should now pass.

src/Makevars
```diff
@@ -1,2 +1,3 @@
 
 CXX_STD = CXX11
+PKG_CPPFLAGS = -mavx -maes
```
eddelbuettel, line 3: I suspect we cannot add those unconditionally as they are likely non-portable, possible even betweem `clang` and `gcc`.

eddelbuettel:
> Do you mean adding the benchmark as a vignette in a separate PR without meow? I would be happy to!

It sounded like you were not sure about meow as a contribution so I offered the benchmark as an alternative.  Happy to do both.

src/Makevars
nx10: Yes exactly, this is what I meant with

> meow does require x64 as well as hardware AES, so I assume it's not a good fit for digest over all.

nx10:
> > Do you mean adding the benchmark as a vignette in a separate PR without meow? I would be happy to!
> 
> It sounded like you were not sure about meow as a contribution so I offered the benchmark as an alternative. Happy to do both.

Yes this is exactly what I meant. Sorry about the confusion.

nx10:
Any thoughts on why spookyhash is so much faster than it should be?

I noticed that spookyhash calls `R_Serialize` from C while the other algorithms call `serialize` from R.

eddelbuettel:
Nope!  Time for some profiling, maybe?

nx10:
I think I found something interesting: 

Serialization memory out streams/connections seem to start with a reserved space of 0 in base R:

https://github.com/wch/r-source/blob/90e365d4f7066b58201275333144be3753835c12/src/main/serialize.c#L2810-L2811

I did a quick test what happens of you preallocate to `object.size()` (with and without padding and reading from the stream):

```R
serialize_fast <- function(object) {
    zz <- rawConnection(raw(object.size(object)-object.size(raw(0))), "r+")
    serialize(object, zz)
    re <- rawConnectionValue(zz)
    close(zz)
    re
}

serialize_fast_pad <- function(object) {
    zz <- rawConnection(raw(object.size(object)-object.size(raw(0))+32), "r+")
    serialize(object, zz)
    re <- rawConnectionValue(zz)
    close(zz)
    re
}

serialize_fast_pad_noread <- function(object) {
    zz <- rawConnection(raw(object.size(object)-object.size(raw(0))+32), "r+")
    serialize(object, zz)
    close(zz)
}


ben2 <- function(object, num_reps) {

    df <- rbenchmark::benchmark(
        "old" = {
            serialize(object, NULL)
        },
        "new" = {
            serialize_fast(object)
        },
        "new_pad" = {
            serialize_fast_pad(object)
        },
        "new_pad_noread" = {
            serialize_fast_pad_noread(object)
        },
        replications = num_reps, order = "elapsed")

    df$bandwidth <- sprintf("%.2f GB/s", object.size(object) * num_reps / 1024^3 / df$elapsed)
    df
}
```

This leads to > 2x - 3x the performance of `serialize()`:

|test           | replications| elapsed| relative| user.self| sys.self| user.child| sys.child|bandwidth |
|:--------------|------------:|-------:|--------:|---------:|--------:|----------:|---------:|:---------|
|new_pad_noread |            1|    0.39|    1.000|      0.30|     0.10|         NA|        NA|2.56 GB/s |
|new_pad        |            1|    0.56|    1.436|      0.40|     0.17|         NA|        NA|1.79 GB/s |
|new            |            1|    0.74|    1.897|      0.58|     0.15|         NA|        NA|1.35 GB/s |
|old            |            1|    1.26|    3.231|      0.81|     0.45|         NA|        NA|0.79 GB/s |

I will try to implement this directly in C next.

nx10:
I have implemented this now in C and did some more benchmarks:

https://github.com/nx10/serialize_prealloc

I think this should be fixed in base R, and I will look into how to submit a PR/patch there for now.

Feel free to close this PR, I will come back to digest soon for the benchmark vignette. 
Thank you for your time!

eddelbuettel:
Nice stuff!  I hope you find an R Core sponsor of that.

Now that leaves us with
- meow: do you think there is/isn't a PR here?
- new vignette on benchmarking: no rush

nx10:
> meow: do you think there is/isn't a PR here?

I don't think it's woth it. xxhash has close to the same speed with AVX2 extensions enabled, and meow does not work at all on other processors (https://github.com/Cyan4973/xxHash/wiki/Performance-comparison) 
So if something really fast is needed in digest, I would recommend looking into conditionally enabling the AVX2 extensions for xxhash. 

Maybe xxhash should also be updated (digest uses version [v0.6.5](https://github.com/eddelbuettel/digest/commit/0c083bcad69c5470a04c7c245d6a38982dbd4f5d) current is [v0.8.1](https://github.com/Cyan4973/xxHash/releases/tag/v0.8.1))

> new vignette on benchmarking: no rush

Thanks! On my to-do list :)

eddelbuettel:
Closing this now.

Issue #185, opened 2023-06-23
AES CBC mode: add support for padding


daattali:
When encrypting or decrypting in CBC mode, the input must have a length of a multiple of 16. This is quite restrictive. Many online tools and other libraries provide an option to pad, such as PKCS. It would be useful if AES provided padding as well.

eddelbuettel:
Sure. Do we like you could propose a PR doing so for AES?

daattali:
I can probably write that , but I'm a noob when it comes to encryption, I didn't know what AES or padding was until a few hours ago. Are there other padding methods, or is the one I mentioned the main one? Do you want that be done automatically if the input length is not a multiple of 16?

eddelbuettel:
AES was a contributed.  I personally do not use it much.  

Don't worry too much about _encryption_ as the `digest` package goes out of its way stating its aim is _hashing_.

daattali:
Actually I won't be adding this, because I'm not even sure the CBC encryption is working correctly. 

I'm getting different results with this package vs {openssl} or online tools. All other tools are in agreement, but digest::AES seems to cut the result too short. Example:

```r
msg <- charToRaw("0123456789012345")
key <- charToRaw("ABCDEFGHIJKLMNOPQRSTUV==")
iv <- charToRaw("61e605afd0fd15f3")

print(as.raw(openssl::aes_cbc_encrypt(data = msg, key = key, iv = iv)))

print(digest::AES(mode = "CBC", key = key, IV = iv)$encrypt(msg))
```

Output:

```
[1] dc 5c d2 89 5d 4d c2 26 2c 5a 2d 58 42 a0 38 29 08 f0 5e c0 37 c3 27 54 50 a9 48 b9 e9 1d 83 c8
[1] dc 5c d2 89 5d 4d c2 26 2c 5a 2d 58 42 a0 38 29
```

I tried other messages and the issue still happens. Perhaps I'm not using the function correctly, but I'm just going to use {openssl} even though it was initially my second choice because this package is so light on dependencies

eddelbuettel:
Whatever works for you. I am fairly certain that what we claim are _reference_ values are in fact reference values.  Eg from the start of the manual page and unit test we have (for FIPS 197, the first one covered)

```r
plaintext       <- hextextToRaw("00112233445566778899aabbccddeeff")

aes128key       <- hextextToRaw("000102030405060708090a0b0c0d0e0f")
[...]

PR #186, opened 2023-06-25
AES: add `padding` argument to allow CBC mode to use padding

daattali:
Soon after writing my last comment in https://github.com/eddelbuettel/digest/issues/185 I realized why my tests were not working. Every other tool I used to encrypt/decrypt using CBC mode is assuming that padding is used. 

Once I implemented padding, digest::AES returned exactly the same as expected

**NOTE:** I did not add an entry in the changelog, did not update the version in DESCRIPTION, and did not update the documentation site files

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-06-25  Dean Attali  <daattali@gmail.com>
+
+        * R/AES.R: Add `padding` parameter to `AES()` to allow CBC mode to use PKCS#7 padding
+
 2023-04-30  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,10 @@
 2023-06-25  Dean Attali  <daattali@gmail.com>
 
-        * R/AES.R: Add `padding` parameter to `AES()` to allow CBC mode to use PKCS#7 padding
-
-        * man/AES.Rd: Document the `raw` argument of `AES()$decrypt()`
+        * R/AES.R: Add `padding` parameter to `AES()` to allow CBC mode to use
+        PKCS#7 padding
+        * inst/tinytest/test_aes.R: Add tests for new `padding` parameter
+        * man/AES.Rd: Document the `raw` argument of `AES()$decrypt()` and the
+        new `padding` parameter
 
 2023-04-30  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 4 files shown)


daattali:
Also note that in my (non-expert) opinion, padding should always be on and it shouldn't even be an option. I say that because it seems that {openssl} and the few online tools I used make that choice. But if I would do that here, it would be a breaking change, so in order to not break anyone's existing code, I added it as a parameter that explicitly needs to be turned on.

eddelbuettel:
Agree that a parameter is good for transition management.  I do appreciate

- entries in Changelog (and often say so in PR comment, so please add one -- format should be clear, and it is two spaces between the three fields)
- not changing the DESCRIPTION and other parts best left ro the person handling the release(s), usually the maintainer

inst/tinytest/test_aes.R
```diff
@@ -117,6 +117,51 @@ aes <- AES(key, mode="CFB", IV=raw(15))
...
+expect_identical(
+  paste(AES(raw_text, "CBC", raw_text, padding = TRUE)$encrypt("testingalongerstring"), collapse = ""),
+  "2d870ada654f4bdff182497a88e08013a232bfb5701c12a4d3b29f4f66e74ecd"
+)
+expect_identical(
+  AES(raw_text, "CBC", raw_text, padding = TRUE)$decrypt(
+    AES(raw_text, "CBC", raw_text, padding = TRUE)$encrypt("testingalongerstring")
+  ),
+  "testingalongerstring"
+)
```
eddelbuettel, line 167: Appreciate the new tests!

eddelbuettel (commented):
That looks good. 

Maybe add a ChangeLog entry?

daattali:
Added changelog entry

ChangeLog
```diff
@@ -1,3 +1,9 @@
+2023-06-25  Dean Attali  <daattali@gmail.com>
+
+        * R/AES.R: Add `padding` parameter to `AES()` to allow CBC mode to use PKCS#7 padding
+
+        * man/AES.Rd: Document the `raw` argument of `AES()$decrypt()`
```
eddelbuettel, line 5: :+1:   

Add a line for added tests?

daattali:
Added the documentation of raw from #187 and updated changelog with all files touched

eddelbuettel:
Very nice -- in it goes.  Thanks for all the detailed follow-ups, much appreciated.

eddelbuettel:
New version is now on CRAN.  Thanks again!

PR #187, opened 2023-06-25
The 'raw' argument (digest()) in aes$decrypt was not documented

daattali:


eddelbuettel:
If find this is described in the `man/digest.Rd`.

daattali:
I'm not talking about the `digest()` function. My original title is correct.

The `AES()` function returns an object that has a `decrypt()` function, which takes a `raw` argument. That argument is currently undocumented. I've added the documentation in the correct place.

Here's a before and after snippet from the documentation of `AES()`:

image

VS

image

eddelbuettel:
But raw is no argument to `AES` or am I undercaffeniated?

```r
AES <- function(key, mode=c("ECB", "CBC", "CFB", "CTR"), IV=NULL) {
```

eddelbuettel:
For simplicity and as we now have two open PRs hitting the same file, how would you feel about cherry-picking this commit into the other open PR?

daattali:
I thought you would prefer to keep the history clear and have each PR be about exactly one thing, not to combine two different things together. Let me know if you do prefer me to close this and add it to the other PR.

eddelbuettel:
Fair point but it also just one PR (and to, issue) of "enhancing/updating digest for AES".  Plus two competing PRs racing the same file is just ... odd.   So if you could a quick cherry-pick of the sole comitt from the other PR (or a merge of its branch or whatever many other methods git has) would be nice.   

This all really helps: you identified a weak spot, and brushed it up.  That's how it is supposed to work, no?   :wink:

PR #191, opened 2023-06-27
Use endian macros to set endianness

barracuda156:
@eddelbuettel Could you please see if this works for you?

These macros should be supported both by GCC and Clang. And if undefined, should use default 0.

### DIFF
--- a/src/crc32c/crc32c_config.h
+++ b/src/crc32c/crc32c_config.h
@@ -6,7 +6,11 @@
 #define CRC32C_CRC32C_CONFIG_H_
 
 // Define to 1 if building for a big-endian platform.
+#if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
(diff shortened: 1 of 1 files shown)


eddelbuettel:
My general preference is to continue a discussion in an issue ticket until a consensus, and only then start PRs.  Oh well.

Now, per https://sourceforge.net/p/predef/wiki/Endianness/ this should work.  _I cannot test this_ because you full well know I do not have big endian metal at my disposal.  Apart from that, yes it "may work".

barracuda156:
(Rebased just to add that it closes the respective issue.)

eddelbuettel:
I made two commits for the related M1 issue this morning.  It would nice if you could rebase (or else I can squash-merge) and I always appreciate a ChangeLog entry (format is two spaces between names, date, email fields).  Can you add that?

barracuda156:
Yes, sure, will do today.

eddelbuettel:
Ping re https://github.com/eddelbuettel/digest/pull/191#issuecomment-1609317986

barracuda156:
@eddelbuettel Thank you!

P. S. Sorry for not being fast enough with rebase.

PR #194, opened 2023-08-03
Add version dependency for tinytest

MichaelChirico:
This is required for `expect_length()`, per the NEWS:

https://github.com/markvanderloo/tinytest/blob/eafa51d509747829273351b8ca61b84c76030124/pkg/NEWS#L13-L14

Given 1.4.1 is less than 6 months old, we might also add a workaround to avoid such a recent package version. Happy to make that change.

### DIFF
--- a/tests/tinytest.R
+++ b/tests/tinytest.R
@@ -12,7 +12,7 @@ if (requireNamespace("tinytest", quietly=TRUE)) {
     if (packageVersion("tinytest") >= "0.9.3") {
         ## expect_length is in tinytest 1.4.1
         if (!"expect_length" %in% getNamespaceExports("tinytest")) {
-            expect_length <- function(x, n) stopifnot(length(x) == n)
+            expect_length <- function(x, n) expect_equal(length(x), n)
         }
         tinytest::test_package("digest")
     }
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Sorry where is it used and which change brought it in?

MichaelChirico:
Linking for reference:

https://github.com/eddelbuettel/digest/blob/d5cc4c798fe1f271a7a02a8526eca7834978df46/inst/tinytest/test_aes.R#L129-L130

eddelbuettel:
> Linking for reference:

Yes thanks found it too both via `grep` / `ag` and `git blame`.  From June it seems.  Fair catch.

tests/tinytest.R
```diff
@@ -12,7 +12,7 @@ if (requireNamespace("tinytest", quietly=TRUE)) {
     if (packageVersion("tinytest") >= "0.9.3") {
         ## expect_length is in tinytest 1.4.1
         if (!"expect_length" %in% getNamespaceExports("tinytest")) {
-            expect_length <- function(x, n) stopifnot(length(x) == n)
+            expect_length <- function(x, n) expect_equal(length(x), n)
```
eddelbuettel, line 15: Good enough for governement work.  There is a bit more in `tinytest::expect_length()` but this may.  Can you actually test this once in a one-off with an ancient `tinytest` ?

MichaelChirico: Good idea -- I'd missed `tinytest` isn't attached. Tests pass.

PR #195, opened 2023-08-03
Missing include for uint32_t

MichaelChirico:
Same provenance as https://github.com/RcppCore/Rcpp/pull/1272.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-08-03  Michael Chirico  <chiricom@google.com>
+
+        * src/digest.c: `#include <stdint.h>` for `uint32_t`
+
 2023-06-28  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Release 0.6.33
--- a/src/digest.c
+++ b/src/digest.c
@@ -21,6 +21,7 @@
 
 */
 
+#include <stdint.h> // for uint32_t
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>


eddelbuettel:
Same rules.  ChangeLog makes it happen more easily.   :grinning:

MichaelChirico:
Sorry, you're too quick 🔫 🤠 🔫 

I'm editing in the web (the `Edit` button, not web VScode) where going one file at a time is slightly easier.

eddelbuettel:
Yes I did that once or twice too and actually had trouble editing a second or third file in the same auto-fork it creates.  Standard fork is just so much easier and better and (ahem) allows you to run tests locally like any responsible committer would :heart_eyes:

MichaelChirico:
Yes, I switch to VScode-on-web for multi-step changes... but running tests locally is easier said than done! Usually easier to let GH CI do the testing when there's compilation involved and I don't have my personal machine handy.

eddelbuettel:
Now you have ChangeLog conflict to resolve, if you could.

MichaelChirico:
I presume this sort of conflict is what Rcpp's .gitattributes is trying to avoid, is that right? I can add the same here if you'd like

https://github.com/RcppCore/Rcpp/blob/ba19cde26d9fbc490e0dcb9ff7b7d148488b2893/.gitattributes#L1

MichaelChirico:
Style question, should there be two ChangeLog entries here, or one entry with two bullet points?

eddelbuettel:
In this (very old) GNU Style I have only ever seen one 'date, author' entry which would then contain multiple bullet points.

You can group them and split them with an empty line between 'topics'.  There should be precedents in the same (20 year old !!) file.

eddelbuettel:
Replaced by #196

PR #205, opened 2024-06-22
blake3: backport upstream patch for Big-endian platforms

barracuda156:
See: https://github.com/BLAKE3-team/BLAKE3/commit/0b13637ae31c2e7e1a471e39258606fabb01685e

@eddelbuettel This fixes `blake3` for Big endian.

As for spooky hash, perhaps it is either to be disabled or left as is: it works on Big-endian, but computes different values, so it will fail comparisons against pre-computed Little-endian values.
See: https://github.com/facebook/folly/blob/4c603f8c2add8d0228de0e073c5ae3ce9b02b6f3/folly/hash/SpookyHashV2.h#L35-L36
Also:
> The hash would run, and it would produce equally good results, but they'd be different results from little-endian platforms.
http://burtleburtle.net/bob/hash/spooky.html

### DIFF
--- a/src/blake3.c
+++ b/src/blake3.c
@@ -89,7 +89,7 @@ INLINE void output_chaining_value(const output_t *self, uint8_t cv[32]) {
   memcpy(cv_words, self->input_cv, 32);
   blake3_compress_in_place(cv_words, self->block, self->block_len,
                            self->counter, self->flags);
-  memcpy(cv, cv_words, 32);
+  store_cv_words(cv, cv_words);
 }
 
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Thanks for helping concretely with a patch, this is appreciated.

barracuda156:
@eddelbuettel Thank you for merging!

eddelbuettel:
My pleasure, that was nicely on point.  

Can you test the new branch / pr #206 where I no longer run tests for 'spookyhash' if on big endian?

PR #209, opened 2024-08-15
use API variants of Calloc / Free

### LINKED ISSUE #208: digest fails to build with recent r-devel builds
The most recent r-devel builds now set `STRICT_R_HEADERS` on by default, which implies that these APIs are no longer available:

https://github.com/wch/r-source/blob/af085e9ea6c901381101afa8e3a8275a2ecdf95d/src/include/R_ext/RS.h#L52-L59

These are currently reported on CRAN under the "Strict" checks, as per https://www.stats.ox.ac.uk/pub/bdr/Strict/digest.out, but recent NEWS entries suggest this will likely be the default in the next release.

Fortunately, the fixes should be straightforward, and I can prepare a PR.

kevinushey:
Closes #208.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2024-08-15  Kevin Ushey  <kevinushey@gmail.com>
+
+	* src/raes.c: Calloc => R_Calloc; Free => R_Free
+
 2024-07-15  Dirk Eddelbuettel  <edd@debian.org>
 
 	* README.md: Switch some URLs from http to https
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -1,3 +1,7 @@
+2024-08-15  Kevin Ushey  <kevinushey@gmail.com>
+
+	* src/raes.c: Calloc => R_Calloc; Free => R_Free
```
eddelbuettel, line 3: I make that -> instead of => but all good.  Thanks for the PR ... and a ChangeLog entry!

eddelbuettel (approved):
:shipit:

PR #210, opened 2024-08-18
expand hashing algorithms that support `raw = TRUE` option

pearsonca:
I would like to use `digest` in some other work, but can't at the moment because the hashing algorithm I'd like to use doesn't support returning raw results.

I've worked up some changes to expand support for `raw = TRUE` for all the algorithms. I've added a test to confirm this is working.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-08-19 Carl A. B. Pearson <carl.ab.pearson@gmail.com>
+
+        * src/digest.c: enable all hashing algorithms to return raw output.
+        * inst/tinytest/test_raw.R: test raw vs not consistency for all algos.
+
 2024-08-15  Kevin Ushey  <kevinushey@gmail.com>
 
 	* src/raes.c: Calloc => R_Calloc; Free => R_Free
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2024-08-19 Carl A. B. Pearson <carl.ab.pearson@gmail.com>
+2024-08-19  Carl A. B. Pearson  <carl.ab.pearson@gmail.com>
 
   * src/digest.c: enable all hashing algorithms to return raw output.
   * inst/tinytest/test_raw.R: test raw vs not consistency for all algos.
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,7 @@ Authors@R: c(person("Dirk", "Eddelbuettel", role = c("aut", "cre"), email = "edd
              person("Michael", "Chirico", role="ctb", comment = c(ORCID = "0000-0003-0787-087X")),
              person("Kevin", "Ushey", role="ctb"),
              person("Carl", "Pearson", role="ctb", comment = c(ORCID = "0000-0003-0701-7860")))
-Version: 0.6.38
+Version: 0.6.37.1
 Date: 2024-08-19
 Title: Create Compact Hash Digests of R Objects
 Description: Implementation of a function 'digest()' for the creation of hash
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,8 +39,9 @@ Authors@R: c(person("Dirk", "Eddelbuettel", role = c("aut", "cre"), email = "edd
              person("Winston", "Chang", role="ctb", comment = c(ORCID = "0000-0002-1576-2126")),
              person("Dean", "Attali", role="ctb", comment = c(ORCID = "0000-0002-5645-3493")),
              person("Michael", "Chirico", role="ctb", comment = c(ORCID = "0000-0003-0787-087X")),
-             person("Kevin", "Ushey", role="ctb"))
-Version: 0.6.37
+             person("Kevin", "Ushey", role="ctb"),
+             person("Carl", "Pearson", role="ctb", comment = c(ORCID = "0000-0003-0701-7860")))
+Version: 0.6.38
 Date: 2024-08-19
 Title: Create Compact Hash Digests of R Objects
 Description: Implementation of a function 'digest()' for the creation of hash
--- a/inst/tinytest/test_raw.R
+++ b/inst/tinytest/test_raw.R
@@ -18,3 +18,15 @@ expect_equal(current, expected)
 
 ## feed raw to sha1() to test sha1.raw() as well
 expect_true(sha1(expected) == "75a2995eeec0fcb5d7fa97c676a37f4e224981a1")
+
+for (algo in c("md5", "sha1", "crc32", "sha256", "sha512",
+               "xxhash32", "xxhash64", "murmur32",
+               "spookyhash", "blake3", "crc32c",
+               "xxh3_64", "xxh3_128")) {
+  digestchar <- digest("The quick brown fox", algo = algo, raw = FALSE)
+  digestraw <- paste0(as.character(
+    digest("The quick brown fox", algo = algo, raw = TRUE)
+  ), collapse = "")
+  
+  expect_equal(digestchar, digestraw, info = sprintf("error in %s", algo))
+}
--- a/inst/tinytest/test_raw.R
+++ b/inst/tinytest/test_raw.R
@@ -19,14 +19,16 @@ expect_equal(current, expected)
 ## feed raw to sha1() to test sha1.raw() as well
 expect_true(sha1(expected) == "75a2995eeec0fcb5d7fa97c676a37f4e224981a1")
 
+
 for (algo in c("md5", "sha1", "crc32", "sha256", "sha512",
                "xxhash32", "xxhash64", "murmur32",
                "spookyhash", "blake3", "crc32c",
                "xxh3_64", "xxh3_128")) {
-  digestchar <- digest("The quick brown fox", algo = algo, raw = FALSE)
-  digestraw <- paste0(as.character(
-    digest("The quick brown fox", algo = algo, raw = TRUE)
-  ), collapse = "")
-  
-  expect_equal(digestchar, digestraw, info = sprintf("error in %s", algo))
+
+    digestchar <- digest("The quick brown fox", algo = algo, raw = FALSE)
+    digestraw <- paste0(as.character(
+        digest("The quick brown fox", algo = algo, raw = TRUE)
+    ), collapse = "")
+
+    expect_equal(digestchar, digestraw, info = sprintf("error in %s", algo))
 }
--- a/src/digest.c
+++ b/src/digest.c
@@ -72,13 +72,65 @@ FILE* open_with_widechar_on_windows(const char* txt) {
     return out;
 }
 
+
+// Also already used in sha2.h
+//
+// We can rely on WORDS_BIGENDIAN only be defined on big endian systems thanks to Rconfig.
+//
+// A number of other #define based tests are in other source files here for different hash
+// algorithm implementations notably crc32c, pmurhash, sha2 and xxhash
+//
+// A small and elegant test is also in package qs based on https://stackoverflow.com/a/1001373
+
+// edd 02 Dec 2013  use Rconfig.h to define BYTE_ORDER, unless already defined
+#ifndef BYTE_ORDER
+// see sha2.c comments, and on the internet at large
+#define LITTLE_ENDIAN 1234
+#define BIG_ENDIAN    4321
+#ifdef WORDS_BIGENDIAN
+#define BYTE_ORDER  BIG_ENDIAN
+#else
+#define BYTE_ORDER  LITTLE_ENDIAN
+#endif
+#endif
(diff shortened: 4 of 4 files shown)


eddelbuettel:
Just got in from traveling, and will review.  Thanks for taking the time to propose an improvement.

eddelbuettel (commented):
This is a pretty large PR I may need some more time.

pearsonca:
Of course - no rush. In the process of adding additional raw outputs, I tried to use a generic approach to the wrap up phase, and ended up applying that more generic approach to all of the hashers.

I moved the endian bits to the top in anticipation of needing them, but I didn't really have a handy way to check if the changes work under different endian-ness platforms. I suspect something can be done with the generic approach macros to make addressing that (outside of the hashing algorithms themselves) will make handling that switch-like.

src/digest.c
```diff
@@ -220,37 +273,43 @@ SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Se
...
+        if (leaveRaw) {
+          rev_memcpy(output, &val.high64, 8);
+          rev_memcpy(output + 8, &val.low64, 8);
+        } else {
+          snprintf(output, 128, "%016" PRIx64 "%016" PRIx64, val.high64, val.low64);
+        }
         break;
     }
     case 101: {     /* md5 file case */
         int j;
```
pearsonca, line 308: missed this one - will prune.

pearsonca: (and some other `int j` vestigal declarations.)

eddelbuettel:
Yes.  We can possibly split this also into more generic 'redoing of things' and a new 'raw export' feature.  

I think none of the CRAN machines are of the other endian type now that the Solaris machine is retired. But we get this when I upload to Debian where I maintain this as a package too.

Also, you may have seen that @kevinushey caught last week that we needed a microtweak for 'STRICT_HEADERS' for one of the imported hasher -- something I missed / ignored as it was third-party code, I use Rf_ prefixed whereever possibly for  many years in my code -- and I just got the CRAN nag about it so I shipped a new 0.6.37 for which I will commit the small remaining changes once it is accepted.  So we probably want a rebase anyway in due course.  I'll ping you again.

src/digest.c
```diff
@@ -72,16 +72,80 @@ FILE* open_with_widechar_on_windows(const char* txt) {
...
+#define dump32(WHAT) decl_tmp32(WHAT);     \
+if (leaveRaw) {                            \
+  rev_memcpy(output, &tmp, output_length);     \
+} else snprintf(output, 128, "%08x", tmp);
+
+#define dump64(WHAT) decl_tmp64(WHAT);     \
+if (leaveRaw) {                            \
+  rev_memcpy(output, &tmp, output_length);                                                     \
+} else snprintf(output, 128, "%016" PRIx64, tmp);
+
```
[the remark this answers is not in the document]
pearsonca: Works for me. Happy to retool them as local functions instead of macros, but that is definitely getting into larger refactor territory.

How about this: keep these changes approximately as-is. Introduce new issue for discussion laying out some refactoring objectives (e.g. the repetitious cases for files vs not also annoyed me), and I'll contribute another PR against that.

Re macros - for some copy pasta, I'd likely use macros in C++ as well, but these are more a byproduct of how I thought through the approach (and then didn't wrap up as functions) rather than being a particularly good answer for this approach.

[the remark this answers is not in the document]
eddelbuettel: Sounds good.  We have a mix of using helpers, and of repetition as the code is (in some places) thirty years old, we now know some things better than we did then ... and also because I was generally really prudent (you could say paranoid) about not changing working code once the package got more widely used.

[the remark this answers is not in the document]
pearsonca: Alright, I will be on the lookout for new issue (feel free to tag / assign me / whatever). I'll discuss there a bit to reach consensus on scope / intent, then make a new PR.

Please let me know here any specific revisions needed for *this* change set here. Alternatively, should be setup for any maintainers of this repo to be able to make changes directly to the source of this PR, so y'all can just make changes directly.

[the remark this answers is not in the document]
pearsonca: WILCO.

Aside: for the standard contributions content, y'all should should consider a CONTRIBUTING.md, explaining those details. Also, introducing issue / PR templates with those items (+whatever else). Pre/auto-nagging is the best kind.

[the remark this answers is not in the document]
eddelbuettel: Sure.  But keep in mind that over the years I added dozens of repos here and it is tedious to copy this around.  I wrote things up once in my most-visible repo [in this file](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) and sometimes refer to it from other repos.  It's mostly about the 'discuss first, pr later' approach that can help (but that I myself also do not _religiously_ follow because somethings small things are just ... small).  [ Need to update that file to w.r.t. to C++ standards. ]

So while point taken, templates can also be heavy for some more marginal repos.  But we do this at work too etc and I know where you are coming from.  We had not needed it here yet.  (Also note above about paranoia / care.  I don't think I bulk formatted here either.)

[ And then as a Wilco (the band) fan and Chicagoan, where did that come from?  Love it.  :heart:  ]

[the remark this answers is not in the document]
eddelbuettel: Release 0.6.37 is now at CRAN so I just pushed. It's mostly DESCRIPTION and the now mandatory Authors@R (to which you should also add yourself, likely in last spot (as the ctb list is roughly chronological)).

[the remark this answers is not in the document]
pearsonca: yeah, heard re maintaining consistent infrastructure docs. github organizations make that relatively easy, with inherit / override as appropriate on a per project basis, but that kind of reorg is its own set of churn.

I'll update the PR to be against 0.6.37 & make the other changes (just did whitespace + ChangeLog, will get the others in a tic).

While I am a fan of Wilco and Chi-town, WILCO is a NATO standard phone talker / radio operator procedure word (short for "understood: will comply"). I also almost wrote just there "proword", which is the shorthand for procedure word.

DESCRIPTION
```diff
@@ -39,8 +39,9 @@ Authors@R: c(person("Dirk", "Eddelbuettel", role = c("aut", "cre"), email = "edd
              person("Winston", "Chang", role="ctb", comment = c(ORCID = "0000-0002-1576-2126")),
              person("Dean", "Attali", role="ctb", comment = c(ORCID = "0000-0002-5645-3493")),
              person("Michael", "Chirico", role="ctb", comment = c(ORCID = "0000-0003-0787-087X")),
-             person("Kevin", "Ushey", role="ctb"))
-Version: 0.6.37
+             person("Kevin", "Ushey", role="ctb"),
+             person("Carl", "Pearson", role="ctb", comment = c(ORCID = "0000-0003-0701-7860")))
+Version: 0.6.38
```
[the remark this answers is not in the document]
eddelbuettel: Please making 0.6.37.1.   It's a bit of my personal tick to leave the determination of what the release is to the maintainer (me) but like you I also _much_ insist of making in-flight versions different from released one.

Been using four-digit 'major.minor.patch.micro' for that since forever.  Works well.

src/digest.c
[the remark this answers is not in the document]
eddelbuettel: Today I learned ...

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2024-08-19 Carl A. B. Pearson <carl.ab.pearson@gmail.com>
```
eddelbuettel, line 1: We want two spaces between the three components.

```suggestion
2024-08-19  Carl A. B. Pearson  <carl.ab.pearson@gmail.com>
```

DESCRIPTION
pearsonca, line 44:
```suggestion
Version: 0.6.37.1
```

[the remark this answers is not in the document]
pearsonca: Done.

src/digest.c
```diff
@@ -72,16 +72,78 @@ FILE* open_with_widechar_on_windows(const char* txt) {
...
+SEXP is_big_endian(void) {
+    return Rf_ScalarLogical(BYTE_ORDER == BIG_ENDIAN);
+}
+
+SEXP is_little_endian(void) {
+    return Rf_ScalarLogical(BYTE_ORDER == LITTLE_ENDIAN);
+}
+
+/* for any algorithm that has output as unsigned char*, can use resolve */
+#define resolve(WHAT) if (leaveRaw) {                          \
```
eddelbuettel, line 106: How about renaming to either `_resolve` (to signal internal function) or maybe state more clearly what it is for, ie something like `_cast_from_unsigned_char` or alike?

src/digest.c
```diff
@@ -72,16 +72,78 @@ FILE* open_with_widechar_on_windows(const char* txt) {
...
+
+/* for any algorithm that has output as an integral value, can use dump */
+
+void rev_memcpy(char *dst, const void *src, int len) {
+    for (int i = 0; i < len; i++) {
+        dst[i] = ((char*)src)[len - i - 1];
+    }
+}
+
+#define dump32(WHAT) decl_tmp32(WHAT);            \
```
eddelbuettel, line 123: Can we use `_dump32` or something more evocative?

src/digest.c
```diff
@@ -72,16 +72,78 @@ FILE* open_with_widechar_on_windows(const char* txt) {
...
+if (leaveRaw) {                                   \
+    rev_memcpy(output, &tmp, output_length);      \
+} else snprintf(output, 128, "%08x", tmp);
+
+#define dump64(WHAT) decl_tmp64(WHAT);            \
```
eddelbuettel, line 128: As above: maybe `_dump64` or something longer?

inst/tinytest/test_raw.R
```diff
@@ -18,3 +18,17 @@ expect_equal(current, expected)
...
+               "xxhash32", "xxhash64", "murmur32",
+               "spookyhash", "blake3", "crc32c",
+               "xxh3_64", "xxh3_128")) {
+
+    digestchar <- digest("The quick brown fox", algo = algo, raw = FALSE)
+    digestraw <- paste0(as.character(
+        digest("The quick brown fox", algo = algo, raw = TRUE)
+    ), collapse = "")
+
+    expect_equal(digestchar, digestraw, info = sprintf("error in %s", algo))
```
eddelbuettel, line 33: :+1:  -- that is a pretty good example.

src/digest.c
```diff
@@ -72,16 +72,78 @@ FILE* open_with_widechar_on_windows(const char* txt) {
...
+    rev_memcpy(output, &tmp, output_length);      \
+} else snprintf(output, 128, "%016" PRIx64, tmp);
+
 SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Seed) {
     size_t BUF_SIZE = 1024;
     FILE *fp=0;
     char *txt;
     int algo = INTEGER_VALUE(Algo);
-    int  length = INTEGER_VALUE(Length);
+    int length = INTEGER_VALUE(Length);
```
eddelbuettel, line 138: Thanks!

src/digest.c
```diff
@@ -116,66 +178,52 @@ SEXP digest(SEXP Txt, SEXP Algo, SEXP Length, SEXP Skip, SEXP Leave_raw, SEXP Se
     switch (algo) {
     case 1: {     /* md5 case */
         md5_context ctx;
-        output_length = 16;
-        unsigned char md5sum[16];
-        int j;
+        output_length = 16; // produces 16*8 = 128 bits
+        unsigned char md5sum[output_length];
```
eddelbuettel, line 182: I really appreciate this and similar cleanups.

src/digest.c
pearsonca: propose: `_store_from_char_ptr`

src/digest.c
pearsonca: propose `_store_from_int32` (and `_store_from_int64` re next comment)

src/digest.c
pearsonca: per previous, `_store_from_int64`

src/digest.c
pearsonca: You're welcome. I *think* these kind of normalization changes are going to point the way to larger scale DRY opportunities, which should then make the other "consider X" issues much easier.

src/digest.c
eddelbuettel: That works.  'store' to me is like committing to disk, so 'return' or 'convert' would work too but all good!

eddelbuettel:
Thanks for all the extended changes, and of course the PR. Had to do a minor cleanup at end because I had committed from issue #211 in the meantime (more or less absent-mindedly).

Issue #212, opened 2024-08-21
digest.c refactoring


pearsonca:
While working #210, I noticed opportunities for opportunites for consolidating / refactoring code in digest.c.

In some sense, digest.c is providing a [Facade](https://en.wikipedia.org/wiki/Facade_pattern) around interacting with the various quirks of the imported hashing libraries. It does some setup / reinterpretation of the input from R, uses a series of cases to handle each of the different algorithms, and then returns a consistent output.

It seems over time those cases have been internally implemented in slightly different ways (and I imagine, the hasher interfaces have changed over time after initial implementation). There is now a fair bit of repetitious code and/or inconsistent naming/etc that could be consolidated. This would make the assorted "Consider X" issues smaller lifts to accomplish, present opportunities for a bit of unit testing on the C side, benchmarking, etc - all the normal benefits of [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself).

I propose the following refactoring objectives:

 - [ ] refactor macros to functions (per discussion in #210)
 - [ ] eliminate the duplicate code for streaming vs not approaches (e.g. by figuring out how to share the setup / return code, while switching between stream vs blob inputs)
 - [ ] more clearly distinguish "jobs" of various functions (e.g. I imagine `digest` would be as a more of traffic manager, with the individual hasher interface logic being modularized out; then adding a new hasher means adding a new consistently-surfaced-function + another case to digest and presto)
 - [ ] additional linting (e.g. spaces around binary operators)

Thoughts? Additional objectives? I think some of these (and possibly additional targets added via discussion) are entangled, but there's some potential to isolate at least some of these as their own PRs - preferences one way or another?

pearsonca:
Can do a PR for just converting macros => functions, sure.

Then maybe draft something for one of the others, and see how it looks w/o (initially) super polishing it? If it seems promising, great, if not, can always toss it.

pearsonca:
Porting discussion from closed PR #218 here:
 - any overhead reductions in digest need maintain code readability
 - better approach may be to introduce no-frills functionality

What could potentially work for that second item:
 - offering a direct (R) interface for each of the algorithms, with minimal parsing / error checking
 - refactor digest to do the error handling etc, then dispatch to the individual mode algorithms
 - longer term that would suggest wanting to have more individualized wrappers in the C layer, but that wouldn't need to happen immediately

pearsonca:
[...]

PR #214, opened 2024-08-22
convert digest macros to function

pearsonca:
Per #212 - converts `_store`... macros to functions.

### DIFF
--- a/src/digest.c
+++ b/src/digest.c
@@ -123,6 +123,7 @@ void _store_from_char_ptr(const unsigned char * hash, unsigned char * const outp
             *outputp++ = sha2_hex_digits[*d & 0x0f];
             d++;
 #else
+            // a char = 2 hex digits => to (0-9A-F)-charset = writing 2 spots
             snprintf(output + j * 2, 3, "%02x", hash[j]);
 #endif
         }
@@ -138,17 +139,14 @@ void rev_memcpy(char *dst, const void *src, int len) {
     }
 }
 
-// n.b. ripe templating if switching to c++
-
+// n.b. ripe templating to e.g. _store_from_integral<> if switching to c++
 void _store_from_int32(const uint32_t hash, char *output, const int leaveRaw) {
     if (leaveRaw) {
         rev_memcpy(output, &hash, sizeof(uint32_t));
     } else snprintf(output, sizeof(uint32_t)*2 + 1, "%08x", hash);
 }
 
-void _store_from_int64(
-    const uint64_t hash, char *output, const int leaveRaw
-) {
+void _store_from_int64(const uint64_t hash, char *output, const int leaveRaw) {
     if (leaveRaw) {
         rev_memcpy(output, &hash, sizeof(uint64_t));
     } else snprintf(output, sizeof(uint64_t)*2 + 1, "%016" PRIx64, hash);
--- a/src/digest.c
+++ b/src/digest.c
@@ -108,8 +108,7 @@ SEXP is_little_endian(void) {
 
 // USESHA512 seems maybe faster? however, more complex, not obviously faster
 void _store_from_char_ptr(const unsigned char * hash, unsigned char * const output,
-                          const size_t output_length, const int leaveRaw
-) {
+                          const size_t output_length, const int leaveRaw) {
     if (leaveRaw) {
         memcpy(output, hash, output_length);
     } else {
--- a/src/digest.c
+++ b/src/digest.c
@@ -103,16 +103,27 @@ SEXP is_little_endian(void) {
 }
 
 /* for any algorithm that has output as unsigned char*, can use _store_from_char_ptr */
-#define _store_from_char_ptr(WHAT) if (leaveRaw) {                          \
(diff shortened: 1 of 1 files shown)


pearsonca:
This can just be done, but I also think there's a way here to pull lines 678-685 into the `_store`...

src/digest.c
```diff
@@ -103,32 +103,52 @@ SEXP is_little_endian(void) {
...
-} else for (int j = 0; j < output_length; j++) {               \
-    snprintf(output + j * 2, 3, "%02x", (WHAT)[j]);            \
-}                                                              \
-
-#define decl_tmp32(WHAT) uint32_t tmp = (WHAT);
-#define decl_tmp64(WHAT) uint64_t tmp = (WHAT);
+void _store_from_char_ptr(
+    const unsigned char *hash, unsigned char *output,
+    const size_t output_length, const int leaveRaw
+) {
```
eddelbuettel, line 109: Could I plead with you to have indentation follow the rest of the file, please?  I.e.

```c++
void _store_from_char_ptr(const unsigned char *hash, unsigned char *output,
                          const size_t output_length, const int leaveRaw) {
     // ...

Comment (pearsonca) [reply src/digest.c:109]: absolutely - my bad, should have caught that was the style choice.

**Comment by eddelbuettel** [comment]
Overall this looks pretty good, and I like that you are incremental.  Maybe we can tag pulling 678-685 in for another round to make things compartmentalized?

I may turn the reverse depdendency machine on, which I haven't for the previous PR yet, just to see where we are.  Off for a morning run now, will look more later.  Thanks for continuing to kick the can down the road!

**Comment by pearsonca** [comment]
> There's also a bit of benchmarking to be done to see if the SHA512 version of conversion to string output is faster than the snprintf version. Either way, the SHA512 output can start using the shared function - question is whether the shared function should do that version or not.

Have poked at this a bit - confirm "classic" vs sha512 conversion to string is identical.

Hashing 10K random alphanumeric "words" (10 alphanumbers each), 100 times: md5 is mean 0.17834 (0.02043231 sd) with classic, 0.16378 (0.01895324 sd) with sha512 conversion, or ~8% reduction, but not a clear difference (due to uncertainty).

Same setup, but with xxhash32 (where the hashing bit should be faster): 0.15089 (0.01595315 sd)  with classic vs 0.15346 (0.01753347 sd) with sha512 - so ~2% increase, but again unclear difference.

I have a general preference for do-things-the-same-way. Vaguely inclined to believe the SHA512 is version is actually faster, so consolidate to that version?

File: src/digest.c (hunk 1 above)
Comment (eddelbuettel) [reply src/digest.c:109]: Thanks. It is the aspect of the 'now common with the kids these days' default choice I like the least as it "blends" function arguments with the function-local variableas as everything is indented the same way.  I find that less readable / imposing extra cognitive load.  Ah well.

**Comment by pearsonca** [comment]
Aside: there seems to be almost no difference between the algoritm choices in terms of effective time as used in R. I suspect hash time would be more apparent hashing big inputs, with better amortization of the shared-R bit, but my particular use case is many-small-inputs.

Might be a case for some S3 specializations?

**Comment by eddelbuettel** [comment]
It's still early morning so I still do quite understand what you are actually after.  In general I am also in favour of consistent approaches especially on our integration side, and extreme care about meddling with upstream code.

**Comment by eddelbuettel** [comment]
In which case I prefer more transparent / more commonly understood code.  `snprintf()` wins here if the performance is a tie.

(Benchmarking is a black hole, of course, as this will vary between compilers, standard library implementations, etc.)

File: src/digest.c [hunk 2]
```diff
@@ -103,32 +103,52 @@ SEXP is_little_endian(void) {
...
+    if (leaveRaw) {
+        memcpy(output, hash, output_length);
+    } else for (int j = 0; j < output_length; j++) {
+// one char is 2 hex digits => converting to (0-9A-F)-chars == writing 2 spots
```
eddelbuettel, line 113: Very helpful comment, also appreciated!.  My editor would align the `//` with code, though :wink:

eddelbuettel (changes requested):
I think this can go in once you align the function signature.

src/digest.c
```diff
@@ -102,49 +102,51 @@ SEXP is_little_endian(void) {
     return Rf_ScalarLogical(BYTE_ORDER == LITTLE_ENDIAN);
 }
 
-/* for any algorithm that has output as unsigned char*, can use _store_from_char_ptr */
-void _store_from_char_ptr(
-    const unsigned char *hash, unsigned char *output,
-    const size_t output_length, const int leaveRaw
+#ifndef USESHA512
+#define USESHA512 0
+#endif
```
eddelbuettel, line 107: Good idea!

eddelbuettel:
Oh, we forgot a ChangeLog entry.  Oh well.  Next PR or just stick one here, I have a pending commit with two more ORCIDs anyway...

pearsonca:
mmmm, something like: `src/digest.c: switch output handling from macro to function; consolidate stringification loop approaches

eddelbuettel:
Done in 0ca2463

PR #218, opened 2024-08-26
digest.R optimization 1: hash algo lookup

pearsonca:
Extracted from #215

### DIFF
--- a/R/digest.R
+++ b/R/digest.R
@@ -37,12 +37,11 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                    errormode=c("stop","warn","silent"),
                    serializeVersion=.getSerializeVersion()) {
 
-    algo <- algo[1]
-    algoint <- algo_int[[algo]]
+    algoint <- algo_int[[algo[1]]]
     errormode <- match.arg(errormode, c("stop", "warn", "silent"))
 
     if (is.null(algoint)) {
-        return(.errorhandler(paste0("Did not understand algo=", algo), mode=errormode))  # #nocov
+        return(.errorhandler(paste0("Did not understand algo=", algo[1]), mode=errormode))  # #nocov
     }
     
     if (is.infinite(length)) {
@@ -54,10 +53,10 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
         file <- TRUE                  	# nocov
     }
 
-    is_streaming_algo <- algo == "spookyhash"
+    is_streaming_algo <- algoint == 9L
 
     if (is_streaming_algo && !serialize) {
-        return(.errorhandler(paste0(algo, " algorithm is not available without serialization."),  # #nocov
+        return(.errorhandler(paste0(algo[1], " algorithm is not available without serialization."),  # #nocov
                       mode=errormode))                                                      # #nocov
     }
 
@@ -87,7 +86,7 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                              mode=errormode))                                               # #nocov
 
     if (file && is_streaming_algo)
-        return(.errorhandler(paste0(algo, " algorithm can not be used with files."),        # #nocov
+        return(.errorhandler(paste0(algo[1], " algorithm can not be used with files."),        # #nocov
                              mode=errormode))                                               # #nocov
 
     ## HB 14 Mar 2007:  null op, only turned to char if alreadt char
@@ -111,7 +110,7 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                      as.integer(skip),
                      as.integer(raw),
                      as.integer(seed))
-    } else if (algo == "spookyhash"){
+    } else if (algoint == 9L){
         # 0s are the seeds. They are included to enable testing against fastdigest.
         val <- paste(.Call(spookydigest_impl, object, skip, 0, 0, serializeVersion, NULL), collapse="")
     }
--- a/R/digest.R
+++ b/R/digest.R
@@ -37,14 +37,14 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                    errormode=c("stop","warn","silent"),
                    serializeVersion=.getSerializeVersion()) {
 
-    # Explicitly specify choices; this is much faster than having match.arg()
-    # infer them from the function's formals.
-    algo <- match.arg(algo, c("md5", "sha1", "crc32", "sha256", "sha512",
-                              "xxhash32", "xxhash64", "murmur32",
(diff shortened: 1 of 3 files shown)


R/digest.R
```diff
@@ -57,8 +57,8 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
     is_streaming_algo <- algo == "spookyhash"
 
     if (is_streaming_algo && !serialize) {
-        .errorhandler(paste0(algo, " algorithm is not available without serialization."),  # #nocov
-                      mode=errormode)                                                      # #nocov
+        return(.errorhandler(paste0(algo, " algorithm is not available without serialization."),  # #nocov
```
pearsonca, line 60: this seemed to just be missing - not sure if there was a specific reason why this one should *not* be a return?

pearsonca:
```r
profvis::profvis(replicate(1000, digest("8675309", algo = "md5", serialize = FALSE, raw = TRUE)))
```

indicated 4 hot spots: (1) converting algo, (2) converting errormode, (3) converting length, and (4) running the algorithm. This addresses (1)

pearsonca:
(also, deferring the ChangeLog update to (3)?)

eddelbuettel:
Let's please get back down to the convention that *every* PR has a ChangeLog entry.  Thank you.

R/digest.R
eddelbuettel: Good catch.  Had to check the file -- that is indeed what is done in a few other invocations. 

:+1:

R/digest.R
```diff
@@ -107,7 +106,7 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
     if (!is_streaming_algo) {
         val <- .Call(digest_impl,
                      object,
-                     as.integer(algoint),
+                     algoint,
                      as.integer(length),
                      as.integer(skip),
                      as.integer(raw),
```
eddelbuettel, line 112: The entire block is actually "wrong" in the sense of when I wrote it I had likely just learned about `.Call()` over `.C()` and the latter requires this -- but the former does not.  

So we could extend this PR (if you're game) to avoiding / minimising the casting of the other arguments length, skip and raw.  (I will have to check later what type of variable is used at the C level.)

pearsonca: I am thinking of my approach to this as being as each optimization feature comes in, remove the cast here. (definitely these casts can go away, as the casting can be handled correctly on the C side by the SEXP interpreters.

pearsonca: which is to say, yes, I'll definitely address these. Question: do you want them addressed in *this* PR or as part of *this series*?

eddelbuettel: We can take it one at a time too.  

Now, out of curiousity, and I buy into the PR just for "cleaner expression of intent" (ie algo choice as int) but what performance gain do you claim this gets, and is that not likely only relevant for very frequent calls with really small payloads?  They exist, but also consider the `fortune` below ...

```r
> fortunes::fortune("save")

Paul Gilbert: [code comparing speed of apply(z,2,sum) vs. rep(1,10000)%*%z)] which seemed completely contrary to all my childhood teachings.
Douglas Bates: Must have had an interesting childhood if you spent it learning about the speeds of various matrix multiplication techniques.
Paul Gilbert: [...] why is apply so slow?
Brian Ripley: 'so slow' sic: what are you going to do in the 7ms you saved?
   -- Paul Gilbert, Douglas Bates, and Brian D. Ripley (discussing 'the incredible lightness of crossprod')
      R-devel (January 2005)

> 
```

R/digest.R
```diff
@@ -54,10 +53,10 @@ digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
         file <- TRUE                  	# nocov
     }
 
-    is_streaming_algo <- algo == "spookyhash"
+    is_streaming_algo <- algoint == 9L
```
eddelbuettel, line 56: That's trickier.  The is time in code execution, and there is clarity in code.  This no longer wins.

(We could be tricky and make it a factor.  Explicit, easy to read, still pass an integer down.)

pearsonca: can do it as algo[1] comparison instead?

R/digest.R
pearsonca: I have yet to systematically look at any "large" payload cases.  For "small" payload cases, profvis indicates that the match.arg lines are equal to or greater than the actual Call itself.

I have checked 1m character objects, and these parts of the R go down to ~10% of the total (instead of >=2/3rd).

R/digest.R
pearsonca: (i *think* the reason the `algo <- algo[1]` reassignment was slowing things down is triggering when R treats function arguments as read-only vs not, and inadvertantly triggering a copy operation?)

pearsonca:
Alright, having sorted out some what's installed where, what session are we in issues:

```r
randomwords <- sample(c(LETTERS, letters, 0:9), 1e6, replace = TRUE)
randomword <- paste0(randomwords, collapse = "")
profvis::profvis(lapply(randomwords, \(w) digest(w, algo = "md5", serialize = FALSE, raw = TRUE)))
profvis::profvis(replicate(1e3, digest(randomword, algo = "md5", serialize = FALSE, raw = TRUE)))
```
The second one (a lot of really little hashes) is interesting (and this is with this revision; prior to this, algo-related lines totalled up to ~what's seen here for errormode)
Screenshot from 2024-08-26 09-47-41

The first one (a few big hashes), not really. the line translating errormode does actually show up, but corresponds to <1% of the sampled run time.

eddelbuettel:
Sorry, been tied up in other things.   Now, I may need to mull this over a little bit more but I have the feeling that we are making 'mostly light and no hear here'. I am not convinced yet we need a change for performance: I can live with time spent on `match.arg()` in a perfectly valid and common code idiom. Also, as noticed, at least one of the suggested changes _reduces_ code clarity in my eyes.

eddelbuettel:
> and that seems to require mostly work on the R overhead

I said that earlier in passing: maybe it is worth trying an alternate, no-frills accessor.  

How the vectorised access was added was quite nice, maybe we could think of something similar here without requiring baby and bathwater approaches.

>  I do think the errormode version of solving the performance problem could just be pushing the match.arg into the error checking function, without loss of code clarity. I can do that as a PR, if you like.

That has merit too.  We would then not spend those microseconds checking an argument we do not use.  

In short, let's try to refocus on a few, clear, simple, unambiguous improvements. 

Oh and another approach we could consider (and something I now know better than when this started) is to export C-level accessors directly to other packages.  Even less overhead.

pearsonca:
Alright, initiated an trim to the errormode checking, since that's just-better.

Re no frills: how about an xxh(...) function along the lines of the `sha()` functionality? The potential problem there is that all the `sha1(...)` functions ultimately pass to digest, and this would be a different approach.

PR #219, opened 2024-08-28
Move errormode check

pearsonca:
As noted in #218 - errormode argument parsing can be moved into the error handling function for an easy optimization.


eddelbuettel (approved):
LGTM!

PR #220, opened 2024-09-18
draft approach to exporting hash methods directly

pearsonca:
This is a rough in idea for how `{digest}` might approach exposing access to the available hashing methods. The rough idea is:

 - want to make the methods directly available, with minimal overhead (which means the user accepting more of the burden of doing checks)
 - ... also for more basic R types directly
 - ... in a way that's easy to maintain / extend
 - ... and that can be incorporated in a backwards compatible way that eliminates code duplication

Marking this as draft, as the idea is to get a feel for the acceptability of this kind of approach before investing a fair chunk of refactor effort.

### DIFF
--- a/src/digest.h
+++ b/src/digest.h
@@ -1,0 +1,90 @@
+/*
+
+  digest -- hash digest functions for R
+
+  Copyright (C) 2003 - 2024  Dirk Eddelbuettel <edd@debian.org>
+
+  This file is part of digest.
+
+  digest is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  digest is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with digest.  If not, see <http://www.gnu.org/licenses/>.
+
+*/
+
+#include <stdint.h> // for uint32_t
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <R.h>
+#include <Rdefines.h>
+#include <Rinternals.h>
+#include <Rversion.h>
+#include <inttypes.h>
+
+#ifdef _WIN32
+#include <Windows.h>
+#endif
+
+// Also already used in sha2.h
+//
+// We can rely on WORDS_BIGENDIAN only be defined on big endian systems thanks to Rconfig.
+//
+// A number of other #define based tests are in other source files here for different hash
+// algorithm implementations notably crc32c, pmurhash, sha2 and xxhash
+//
+// A small and elegant test is also in package qs based on https://stackoverflow.com/a/1001373
+
+// edd 02 Dec 2013  use Rconfig.h to define BYTE_ORDER, unless already defined
+#ifndef BYTE_ORDER
+// see sha2.c comments, and on the internet at large
+#define LITTLE_ENDIAN 1234
+#define BIG_ENDIAN    4321
+#ifdef WORDS_BIGENDIAN
+#define BYTE_ORDER  BIG_ENDIAN
+#else
+#define BYTE_ORDER  LITTLE_ENDIAN
+#endif
+#endif
+
+SEXP is_big_endian(void);
+SEXP is_little_endian(void);
+
+#if defined(R_VERSION) && R_VERSION >= R_Version(3,0,0)
+#define RVLENGTH(ROBJ) XLENGTH(ROBJ)
+#else
+#define RVLENGTH(ROBJ) LENGTH(ROBJ)
+#endif
+
+#ifndef USESHA512
+#define USESHA512 0
+#endif
+
+#if USESHA512
+static const char *sha2_hex_digits = "0123456789abcdef";
+#endif
+
+// direct_ALGO
+// @param ALGO which algorithm 
+//
+// creates a function taking arguments:
+// @param inputtype: the type of the input, must be one of the enumerated types
+// @param input: the input to hash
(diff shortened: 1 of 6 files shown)


pearsonca:
N.B.: I'm not super-pumped about the macro-based approach. I like it conceptually, but the actual language support is dicey in C and macros seem like the best (of bad) ways to do it - this sort of approach is better in C++ and is more manageable with the linguistic support for classes / templates / etc.

Introducing C++ seems like a heavier overall lift in terms of refactor, however.

src/digest.h
```diff
@@ -0,0 +1,90 @@
...
+#endif
+
+SEXP is_big_endian(void);
+SEXP is_little_endian(void);
+
+#if defined(R_VERSION) && R_VERSION >= R_Version(3,0,0)
+#define RVLENGTH(ROBJ) XLENGTH(ROBJ)
+#else
+#define RVLENGTH(ROBJ) LENGTH(ROBJ)
+#endif
```
eddelbuettel, line 66: Nobody will compile on R older than 3.0.0.  We could just make it an error on config.

And then stick with `XLENGTH` which is known.

src/init.c
```diff
@@ -24,9 +24,15 @@
...
 #include "pmurhash.h"
+#include "digest.h"
+
+#define stringify(TAR) #TAR
+#define Register_ALGO(ALGO) R_RegisterCCallable("digest", stringify(direct_ ## ALGO), (DL_FUNC) &direct_ ## ALGO);
 
 void R_init_digest(DllInfo *info) {
     R_RegisterCCallable("digest", "PMurHash32", (DL_FUNC) &PMurHash32);
+    Register_ALGO(SHA1);
+    Register_ALGO(MD5);
```
eddelbuettel, line 35: Not a huge fan _ex ante_ of `Register_ALGO`.  What `R_init_$PACKAGAME` does is pretty well understood and even if lining up multiple `R_RegisterCCallable()` looks tedious it is _simple_ and a good edtor can help filling in the stanza.  Every other package I have every seen writes it out, there is benefit in seeing the familar `R_RegisterCCallable()` rather than 'what is this ?' one has to chase to its definition to unmask.  And we probably won't expose a many dozen algos.

eddelbuettel:
Thanks for picking this back up. I have the ability to access C-level functions in a packages directly from other packages in a few places so this is surely something we can add.  This PR is already a little on the long side so let's if we can it simply and in stages.  

I put two really quick comments in but I need to come back to this with a bit more time.

src/digest.h
pearsonca: works for me. what's the cleanest way to do that? seems like something to just specify in DESCRIPTION, right?

src/init.c
pearsonca: fair - i'm mostly thinking of this as part of some machinery where the algos get enumerated in one place, then that machinery builds up the appropriate exports (among other things).

an alternative less-magic approach which ensures nothing missed would be some checker that confirms all the algos actually appear here.

src/digest.h
eddelbuettel: Yes.  

Find any comparison for R_VERSION that are that ancient (as here R 3.0.0), default to the 'newer' case and set then in DESCRIPTION set 'R (>= 3.0.0)'.   [ Goes checking. ] Looks like that was the only one in src/.

eddelbuettel:
That bulleted list was helpful. I had not clicked why we needed `digest.h`.  What you wrote makes sense.

What were you thinking in terms of minimal input?  Don't we always start from raw vector we get from the serializating function?  Wouldn't that make it type-agnostic as we don't S3 dispatch in C code?  Or do you want the S3 dispatchers each pick up a matching C function?  That sounds ... like a lot more work.

pearsonca:
> What were you thinking in terms of minimal input? Don't we always start from raw vector we get from the serializating function? Wouldn't that make it type-agnostic as we don't S3 dispatch in C code? Or do you want the S3 dispatchers each pick up a matching C function? That sounds ... like a lot more work.

I think the serialization is a potential opportunity to reduce overhead, but could be wrong! Seems like it should be cheaper to use an int* directly cast as an unsigned char* than to serialize the SEXP contents => use that unsigned char*. But if serialization is 1) doable on the C side (seems likely) and 2) small incremental benefit for the raw types (more skeptical on this), then the reduced engineering + portability gain seems a reasonable tradeoff.

eddelbuettel:
See package RApiSerialize which I use in package RcppRedis to turn _any_ R expression, variable, ... handed down as SEXP into raw bytes (which Redis stores, we would hash them).  Not going through the R side for two calls should help.  We could cover the most common parameterisation as a start.

pearsonca:
Got it, will have a look - but so should expect just shifting serialization to the C side will yield useful gain? Definitely much easier engineering lift.

eddelbuettel:
Well that is what I was thinking we should have a 'demo client' too.  :grinning:   But as I understand it your use case is (many, small, frequent) calls to digest _from the level of C code_ ?  In that case we want the 'digest me this "thing"' approach.

It remains to be seen if we then need some of the convenience of Rcpp for conversion from C/C++ types to SEXP.  For now, and for (say) and int or double vector we can do by hand, no?

pearsonca:
Roughly correct re my use case of interest, but the library I'm building also provides flexibility to hash somewhat arbitrary (though always small) objects. The users should be providing a small distinguishing key, but that needn't be a series of integers - but one choice they should be able to make to improve how the library works for them is to switch to (e.g.) using a series of integers.

eddelbuettel:
"Objects" are still SEXP-representable.  If you can call `digest::digest())` on it we should be able to do it at the C level too, with fewer checks and faster.  Or maybe I miss something?

pearsonca:
> "Objects" are still SEXP-representable. If you can call `digest::digest())` on it we should be able to do it at the C level too, with fewer checks and faster. Or maybe I miss something?

I think that's right, yes.

Alright, I'm going to proceed a bit on the bullet list above, with the tweaks subsequently discussed. That'll come down in separate PRs, but going to leave this one drafted until that series is resolved.

PR #221, opened 2024-09-19
refactor to split header off digest

pearsonca:
Per the title, splits out header. Also cleans up imports and eliminates the version dependency check (already enforced via DESCRIPTION).


pearsonca:
Gah, forgot the changelog update - will add in a min.

eddelbuettel (approved):
Nice and clean

PR #223, opened 2024-12-31
Update spelling typos

billdenney:


src/zconf.h
```diff
@@ -1,5 +1,5 @@
 /* zconf.h -- configuration of the zlib compression library
- * Copyright (C) 1995-2004 Jean-loup Gailly.
+ * Copyright (C) 1995-2004 Jean-Loup Gailly.
```
eddelbuettel, line 2: Just today I came across a reference that this routine is likely one of the three most-widely-used pieces of code in trillions of instances across the globe.

In each and every copyright header Mr Gailly writes his name as 'Jean-loup'. I think we should let him.

src/zutil.h
```diff
@@ -1,5 +1,5 @@
 /* zutil.h -- internal interface and configuration of the compression library
- * Copyright (C) 1995-2003 Jean-loup Gailly.
+ * Copyright (C) 1995-2003 Jean-Loup Gailly.
```
eddelbuettel, line 2: See above.

src/zlib.h
```diff
@@ -1,7 +1,7 @@
 /* zlib.h -- interface of the 'zlib' general purpose compression library
   version 1.2.2, October 3rd, 2004
 
-  Copyright (C) 1995-2004 Jean-loup Gailly and Mark Adler
+  Copyright (C) 1995-2004 Jean-Loup Gailly and Mark Adler
```
eddelbuettel, line 4: See above

src/zlib.h
```diff
@@ -19,7 +19,7 @@
      misrepresented as being the original software.
   3. This notice may not be removed or altered from any source distribution.
 
-  Jean-loup Gailly        Mark Adler
+  Jean-Loup Gailly        Mark Adler
```
eddelbuettel, line 22: See above

man/digest.Rd
```diff
@@ -84,7 +84,7 @@ digest(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
 
   For md5, sha-1 and sha-256, this R implementation relies on standalone
   implementations in C by Christophe Devine. For crc32, code from the
-  zlib library by Jean-loup Gailly and Mark Adler is used.
+  zlib library by Jean-Loup Gailly and Mark Adler is used.
```
eddelbuettel, line 87: See elsewhere.

eddelbuettel:
Hi @billdenney and Happy New Year!  I really appreciate helpful 'eagle eyes' spotting typos would I would like to ask to drastically reduce the scope of the PR.  Everything below `docs/` is auto generated, and I would prefer it continued to be generated by the 'frozen in a docker container' version of material for mkdocs I have here.   So can we regroup, fix what is in 
- DESCRIPTION
- README.md
- man/ (but see digest.Rd comment)
- vignettes

and leaving alone
- docs/ (generated)
- inst/ (not asked for)
- src/ (see comments)

Thanks so much.

eddelbuettel (changes requested):
This won't work as is, please see comments.

Appreciate the PR but this went a little too far, too fast, without enough checking.

src/zconf.h
billdenney: I didn't know that he spelled his name that way. The only reason I made the change is because it was inconsistently capitalized in other places. I agree that anyone should get to spell their own name the way that they like. I've reverted throughout where I made it upper-case, and I made it lower case where it was upper case in a few places.

billdenney:
I have reverted and revised. Now "loup" is always lower case (there were a few places where it was upper-case).

eddelbuettel:
> I have reverted and revised. Now "loup" is always lower case (there were a few places where it was upper-case).

My preference would be to leave the as-is unless you have a *strong* reason to fix.  I.e. when you see me misspell 'numerically' in numerous ways, fix.  How someone else wrote their copyright header is not something I as maintainer support altering---unless _they_ ask me.  So I think we should revert all such changes in the spelling of his name to what it was.  Would that be acceptable?

I once added a spelling list to one package, it mostly got stale. I have no intention of adding one.  If you happen to use other tools by all means do so -- but make use of both `.gitignore` and `.Rbuildignore` as needed, preferably just locally.  Maybe just maybe overzealous use of such tools something to do with fixing in files in that are autogenerated :wink:

man/digest.Rd
```diff
@@ -155,7 +155,7 @@ digest(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
   Antoine Lucas for the integration of crc32; Jarek Tuszynski for the
   file-based operations; Henrik Bengtsson and Simon Urbanek for improved
   serialization patches; Christophe Devine for the hash function
-  implementations for sha-1, sha-256 and md5; Jean-Loup Gailly and Mark Adler
+  implementations for sha-1, sha-256 and md5; Jean-loup Gailly and Mark Adler
```
eddelbuettel, line 158: That's a good fix, correcting my own inadvertent mis-caseing!

eddelbuettel:
The now remaining fixes are great!  Thanks a bunch for these.

eddelbuettel (approved):
Thanks very much!

PR #226, opened 2025-01-01
Add update `sha1.default()` to work with any class and remove what are now duplicate classes

### LINKED ISSUE #224: Feature request: `sha1()` methods for `<-` and `environment` objects
Would you be open to a PR for `sha1.<-()` and `sha1.environment()`?

I realize that `<-` is simple while environments can get very complicated quickly. I think that `<-` would replicate `sha1.call`. It's less clear to me how to best handle `sha1.environment()`.

``` r
library(digest)
sha1(str2lang("a <- b"))
#> Error: sha1() has no method for the '<-' class
sha1(new.env())
#> Error: sha1() has no method for the 'environment' class
```

<sup>Created on 2025-01-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

billdenney:
Fix #224


ChangeLog
```diff
@@ -1,3 +1,7 @@
+2025-01-01  Bill Denney  <wdenney@humanpredictions.com>
+
+  * R/sha1.R: Default sha1 method now works for all classes by default
```
eddelbuettel, line 3: Tab or eight spaces before *

R/sha1.R
```diff
@@ -306,13 +301,6 @@ sha1_attr_digest <- function(x, digits = 14L, zapsmall = 7L, ..., algo = "sha1")
     digest(x, algo = algo)
 }
 
-sha1.call <- function(...) {sha1_attr_digest(...)}
-sha1.character <- function(...) {sha1_attr_digest(...)}
-sha1.factor <- function(...) {sha1_attr_digest(...)}
-sha1.logical <- function(...) {sha1_attr_digest(...)}
-sha1.integer <- function(...) {sha1_attr_digest(...)}
-sha1.raw <- function(...) {sha1_attr_digest(...)}
-
```
eddelbuettel, line 315: Part of me likes this: smaller surface.  Part of me is horrified -- aren't we removing existing interface?

ChangeLog
```diff
@@ -1,6 +1,8 @@
 2025-01-01  Bill Denney  <wdenney@humanpredictions.com>
 
-  * R/sha1.R: Default sha1 method now works for all classes by default
+	* R/sha1.R: Default sha1 method now works for all classes by default. If you
```
eddelbuettel, line 3: s/all/more/ ?  Do we know it's all?

I'd remove the "If you ..." letter to the reader.  Not what we commonly put in ChangeLog :)  Minor style issue.

eddelbuettel:
Reverse-dependency checks came and went.  (The machine I run these on is not often used for `digest` and specific per-package dependencies do not autoinstall so a fair number failed for 'dependency xyz missing' so I ran another baseline check with `digest` from the main branch.  Came out the the same so "no change for worse".)  While that is good I am still not a fan of the removal of the (arguably redundant) S3 accessors.  How would you feel about reverting that part of the change?

billdenney:
I reverted those removals.

eddelbuettel:
Appreciate the reversal.  I keep pondering this, and the actual hard error in there with `stop()` was part of the original support, and I think it makes sense --- it indicates that the S3 support is complete, just as it does in #224.  So a more minimal PR may be my current preference.

I updated my branch, borrowed you additional tests and am running it in reverse dependency checks now.

billdenney:
Is there a scenario where using `sha1_attr_digest()` for classes not specifically handled would be a problem? I ask because I discovered that `<-` and `environment` were missing when I tried to implement an `sha1()` method in the `rxode2` library.

Since `<-` and `environment` are base R classes, it seemed better to put the methods here rather than into the `rxode2` library. Enabling the `sha1.default()` function to handle any class by default via `sha1_attr_digest()` would prevent needing to update `digest` when new classes are necessary that don't fit within another library (as with `<-` and `environment` here).

Another example that would be fixed with a default method that provides a result is `logLik` (found just after I opened #224):

``` r
sha1(logLik(lmx <- lm(x ~ 1, data = data.frame(x = 1:5))))
#> Error in sha1(logLik(lmx <- lm(x ~ 1, data = data.frame(x = 1:5)))): could not find function "sha1"
```

<sup>Created on 2025-01-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

My concern is that this becomes whack-a-mole when going through all the required and recommended libraries that are unlikely to provide their own `sha1()` S3 methods.

eddelbuettel:
I think I prefer to follow the explicit S3 dispatch _where we already use it_ that is in `sha1`.  As demonstrated earlier, if `sha1(foo)` bites for whatever `foo` you have `digest(foo, algo="sha1")` should always be available

```r
> library(digest)
> digest(logLik(lmx <- lm(x ~ 1, data = data.frame(x = 1:5))), algo="sha1")
[1] "e2f78853387414a8b41d7793383ab062aa27241e"
> 
```

meaning that if you need it you could (if push came to shove) define a local sha1 override for it.

The package is 20+ years old, the S3-based sha1 support is nine years old.  I think on balance I rather stay with the design that worked than to (more radically than we need to) alter it now.  The aim is to have something reliable for all/most use cases.

eddelbuettel:
That said, whackamole is a risk, and `logLik` is a good example.  There are a very large number of S3 classes out there, and we cannot possibly cover all.  Then again `digest()` always works on "anything".

Maybe fusing our two competing PR and turning the stop() into a warning and proceeding is a good middle ground.

eddelbuettel:
`message()` it is:

```r
> library(digest)
> sha1(logLik(lmx <- lm(x ~ 1, data = data.frame(x = 1:5))))
sha1() has no method for the 'logLik' class, so using fallback.
[1] "7d7ba7a23dbe3ccc8342811bf6dce184501dc12a"
> suppressMessages(sha1(logLik(lmx <- lm(x ~ 1, data = data.frame(x = 1:5)))))
[1] "7d7ba7a23dbe3ccc8342811bf6dce184501dc12a"
> 
```

Committed in 7af4aa7.  Thoughts?

billdenney:
I like that compromise. I've merged the branches.

eddelbuettel (approved):
Nice. Some procrastination and noodling over alternatives made this better.

eddelbuettel:
Should have altered the PR title (but then did so in an `--amend` commit).  All good now.  

Thanks for your attention to detail and patience on this.

PR #227, opened 2025-01-06
Switch to warning on unknown class in sha, document option

eddelbuettel:
Adding link to https://github.com/inbo/n2kanalysis/blob/main/R/sha1.R showing how to add custom S3 dispatcher
Also update tests as needed

Tagging @ThierryO and @billdenney for a quick review / thumps up-or-down.  This follows yet more discussion in #224.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,14 @@
+2025-01-07  Thierry Onkelinx <thierry.onkelinx@inbo.be>
+
+	* R/sha1.R (sha1.default): Tweak message
+	* man/sha1.R: Extend fallback discussion
+
+2025-01-06  Dirk Eddelbuettel  <edd@debian.org>
+
+	* R/sha1.R (sha1.default): Revert to warning
+	* man/sha1.R: Document fallback, provide reference to package
+	implementing local S3 dispatch override methods
+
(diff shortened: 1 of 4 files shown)


billdenney:
Overall, it looks good.

Stylistically, I would keep the `pattern` with `expect_warning()` so that we know the warning is the one we are adding rather than some other warning from a different place.

eddelbuettel:
> Stylistically, I would keep the pattern with `expect_warning()` 

I kinda hate playing cat-and-mouse but you twisted my arm.  Comitted.

billdenney:
With PKNCA, I was able to catch a bug in a dependency before that dependency went to CRAN with a warning message check. I use them all the time!

ThierryO (approved):
Maybe let the warning suggest to create a user defined method and point to the help file? I would encourage users to do so, especially when they use `sha1()` in their package.

ThierryO:
I've added a suggestion in #228

eddelbuettel:
It's not bad. I am not sure we neeeded another PR / branch for that given that we are talking here ...

Do you have access to a remining 32 bit system?

eddelbuettel:
Anyway, time to move on.  Will merge, edit and merge.

PR #235, opened 2026-07-06
Add Fletcher-32 checksum (algo="fletcher")

chrislongros:
Adds Fletcher's 32-bit checksum as `algo="fletcher"`, alongside `crc32`/`crc32c`. It computes two running sums over 16-bit little-endian words, each modulo 65535, returned as an eight-character hex string:

```r
digest("abcde", algo = "fletcher", serialize = FALSE)   # "f04fc729"
```

Registered as algorithm 14 in `algo_int()`, wired into `digest()` and `getVDigest()`, and documented in `man/digest.Rd`. 

Tests in `inst/tinytest/test_fletcher.R` cover the reference vectors plus the empty, raw, file, and `getVDigest` paths.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2026-07-07  Christos Longros  <chris.longros@gmail.com>
+
+	* src/digest.c: Add Fletcher-32 checksum as algorithm 14
+	* R/digest.R (digest): Register "fletcher" in algo choices
+	* R/vdigest.R (getVDigest): Idem
+	* man/digest.Rd: Document "fletcher" and add reference
+	* inst/tinytest/test_fletcher.R: Add reference check vectors
+
 2025-12-05  Dirk Eddelbuettel  <edd@debian.org>
 
 	* .github/workflows/ci.yaml: Switch to actions/checkout@v6
--- a/R/digest.R
+++ b/R/digest.R
@@ -27,7 +27,7 @@
 digest <- function(object, algo=c("md5", "sha1", "crc32", "sha256", "sha512",
                                   "xxhash32", "xxhash64", "murmur32",
                                   "spookyhash", "blake3", "crc32c",
-                                  "xxh3_64", "xxh3_128"),
+                                  "xxh3_64", "xxh3_128", "fletcher"),
                    serialize=TRUE,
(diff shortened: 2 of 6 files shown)


eddelbuettel:
That looks quite nice and clean, thank you. Will take a closer look in a bit.

Is this something that would misbehave if we had a big-endian machine?  Also, is there is source reference for this?  

[ Goes looking ]

I guess https://en.wikipedia.org/wiki/Fletcher%27s_checksum would suit. Something to add to `digest.Rd` ?

chrislongros:
> That looks quite nice and clean, thank you. Will take a closer look in a bit.
> 
> Is this something that would misbehave if we had a big-endian machine?  Also, is there is source reference for this?  
> 
> [ Goes looking ]
> 
> I guess https://en.wikipedia.org/wiki/Fletcher%27s_checksum would suit. Something to add to `digest.Rd` ?

This implementation is not affected by endianness. I can add a reference to the algorithm in .Rd

eddelbuettel (changes requested):
This looks good to me!  Happy to merge.  Could you do two more things:

1. Add a reference to the method to the main Rd page? Link to wikipedia page is fine.

2. Can you add to the ChangeLog file describing the changes briefly, similar to what you see in file?  The leading line has two spaces between date, name, email.

eddelbuettel (approved):
Thanks for making those additions, and of course for the contribution.

eddelbuettel:
Just noticed at r-universe that the source package had a small slip-up, we overlooked the '%27' for the apostrophe needed an escape.  My bad, should have caught that in code review. Fixing it now.

eddelbuettel:
I also added you as a contributor.  If you have an ORCiD I can add if you drop it here, if you don't no worries.

chrislongros:
> I also added you as a contributor. If you have an ORCiD I can add if you drop it here, if you don't no worries.

Thank you for crediting me as contributor! My ORCID is 0009-0001-2717-0857.

eddelbuettel:
LOL. I am still asleep at the wheel. We both also forgot to run `R CMD check` -- the list of known digest algos needs updating the vigest Rd file too.  I just took care of it.

PR #236, opened 2026-07-10
Add fletcher to README algorithm list and update contributors

chrislongros:
Two small README updates: add fletcher to the algorithm list in the overview, and add myself to the co-authors.


eddelbuettel (approved):
LGTM

eddelbuettel:
Needed a ChangeLog entry too (a little less urgently than the main PR but still...) so I just took care of that.

PR #237, opened 2026-07-11
Add sha256 known-answer tests from the NIST FIPS 180-4 vectors

chrislongros:
Adds known-answer tests for sha256 based on the **NIST FIPS 180-4** vectors (empty string, "abc", the two-block message, one million 'a') via digest(), getVDigest(), and raw=TRUE.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2026-07-11  Christos Longros  <chris.longros@gmail.com>
+
+	* DESCRIPTION (Version, Date): Roll micro version and date
+
+	* inst/tinytest/test_digest.R: Add sha256 known-answer tests using
+	the NIST test vectors for FIPS 180-4
+
 2026-07-10  Christos Longros  <chris.longros@gmail.com>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -1,3 +1,10 @@
+2026-07-11  Christos Longros  <chris.longros@gmail.com>
+
+	* DESCRIPTION (Version, Date): Roll micro version and date
```
eddelbuettel, line 3: Just as a comment, it's probably best "in general" to leave that to the package maintainer. Here it's fine because this is very incremental to a very recent version.

eddelbuettel (approved):
Nice.  It's good to have reference vectors (and I was wondering about that for 'fletcher' too but I presume there is none).

eddelbuettel:
I also noticed we hadn't added 'fletcher' to DESCRIPTION::Description so I just took care of that.

chrislongros:
> Nice.  It's good to have reference vectors (and I was wondering about that for 'fletcher' too but I presume there is none).
> 
> 

I could not find any official test vectors based on a standard. I used the Wikipedia's examples.

eddelbuettel:
Yes, been there / done that. Sometimes a Python (or alike) implementation can serve. True reference values are a gold standard so well played with this PR.

PR #238, opened 2026-07-12
Add sha512 known-answer tests from the NIST FIPS 180-4 vectors

chrislongros:
Adds known-answer tests for sha512 based on the NIST FIPS 180-4 vectors ("abc", the two-block message, one million 'a') via digest(), getVDigest(), and raw=TRUE.


eddelbuettel (approved):
Thanks!

eddelbuettel:
(ChangeLog as always ... I'll will edit it in.)

